PR #5, opened 2025-04-15
Remove obsolete contact address from CoC

grinapo:
Removing old contact data.

Reordering repeated sexuality related attributes.

Adding a sentence about retaliatory actions, which will be ignored the same way as others have been, but, anyway.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -6,9 +6,9 @@
 We as members, contributors, and leaders pledge to make participation in our
 community a harassment-free experience for everyone, regardless of age, body
 size, visible or invisible disability, ethnicity, sex characteristics, gender
-identity and expression, level of experience, education, socio-economic status,
-nationality, personal appearance, race, caste, color, religion, or sexual
-identity and orientation.
+identity and expression, sexual identity and orientation, level of experience, 
+education, socio-economic status, nationality, personal appearance, race, 
+caste, color or religion.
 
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -6,9 +6,9 @@
 We as members, contributors, and leaders pledge to make participation in our
 community a harassment-free experience for everyone, regardless of age, body
 size, visible or invisible disability, ethnicity, sex characteristics, gender
-identity and expression, level of experience, education, socio-economic status,
-nationality, personal appearance, race, caste, color, religion, or sexual
-identity and orientation.
+identity and expression, sexual identity and orientation, level of experience,
```
NinekoTheCat, line 9: > sexual identity and orientation

why is there an and here? isn't a comma more appropriate as these are separate elements of this list?

NinekoTheCat (changes requested):
Overall this is a good pull request, but I think that the one comment I left should be resolved and the title should be changed to mention the changes at the top of the code of conduct as it might be misleading.

Also please remove the comment about
> which will be ignored the same way as others have been, but, anyway.

I don't think it is appropriate.

CODE_OF_CONDUCT.md
Samonitari: Although not a native English speaker myself, I think it's quite idiomatic to compress lists' adjacent elements with a common adjective this way.

PR #37, opened 2025-05-06
Add admin support for LDAP login

### LINKED ISSUE #31: Grant admin privileges via LDAP
When we query for LDAP users on login we can determine whether admin access should be granted.

If the LDAP experts can note what the search results would look like for admins in their org I'd appreciate it.

RatCornu:
Closes #31.

Done the major logic part, I just need now help on how to implement the `demote_admin` function.

Moreover, I added the support of the `{username}` variable replacing in `filter` and `admin_filter` (like it can be found on Jellyfin for example).


jevolk:
Thank you 🙏🏻 I will have a look at this as soon as I'm through with the new release pipeline.

Issue #64, opened 2025-06-14
My updated NixOS build is failing


jevolk:
I have updated the nix materials now on [this branch](https://github.com/matrix-construct/tuwunel/tree/nix) but the build fails moving from june's rocksdb to my rocksdb. We can't use june's rocksdb because it doesn't link with our newer version of rust-rocksdb, so we have to be using my rocksdb now.

![Image](https://github.com/user-attachments/assets/5221a128-5bbb-4cec-8627-9bb3a8bb1146)

Folly should not be involved with our build. I have tried a few cmake flags near where there other ones are but nothing seems to work. I don't know if there is some "upstream derivation" where this is coming from or how this all works. Any help appreciated 🥺

wkordalski:
Hi,

The `rocksdb` derivation is defined in [Tuwunel's `flake.nix` in lines 54-104](https://github.com/matrix-construct/tuwunel/blob/5c43a4f066df2ad010d6e8afb4ccbe211168863d/flake.nix#L54-L104). It overrides some build parameters of [`rocksdb` derivation from Nixpkgs](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/ro/rocksdb/package.nix).

The "upstream derivation" (from Nixpkgs) runs the failing `sed` command in patch phase (and especially in ["postPatch"](https://github.com/NixOS/nixpkgs/blob/ee930f9755f58096ac6e8ca94a1887e0534e2d81/pkgs/by-name/ro/rocksdb/package.nix#L83-L96)):

```nix
postPatch =
  lib.optionalString (lib.versionOlder finalAttrs.version "8") ''
    # Fix gcc-13 build failures due to missing <cstdint> and
    # <system_error> includes, fixed upstyream sice 8.x
    sed -e '1i #include <cstdint>' -i db/compaction/compaction_iteration_stats.h
    sed -e '1i #include <cstdint>' -i table/block_based/data_block_hash_index.h
    sed -e '1i #include <cstdint>' -i util/string_util.h
    sed -e '1i #include <cstdint>' -i include/rocksdb/utilities/checkpoint.h
  ''
  + lib.optionalString (lib.versionOlder finalAttrs.version "7") ''
    # Fix gcc-13 build failures due to missing <cstdint> and
    # <system_error> includes, fixed upstyream sice 7.x
    sed -e '1i #include <system_error>' -i third-party/folly/folly/synchronization/detail/ProxyLockable-inl.h
  '';
```

[...]

PR #65, opened 2025-06-14
Fix building RocksDB on Nix

wkordalski:
Fixes: #64

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -53,7 +53,7 @@
         };
         rocksdb = (pkgs.rocksdb.override {
           liburing = self.liburing;
-        }).overrideAttrs (old: {
+        }).overrideAttrs (final: old: {
           src = inputs.rocksdb;
           version = pkgs.lib.removePrefix
             "v"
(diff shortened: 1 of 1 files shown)


coolGi69:
Can't seem to compile with `nix build github:wkordalski/tuwunel/nix` as it causes the following error in `rocksdb-tuwunel-changes`:
```
   Compiling tuwunel_macros v1.0.0 (/build/source/src/macros)
error[E0658]: use of unstable library feature `inherent_str_constructors`
  --> src/macros/git.rs:56:4
   |
56 |             str::from_utf8(&output.stdout)
   |             ^^^^^^^^^^^^^^
   |
   = note: see issue #131114 <https://github.com/rust-lang/rust/issues/131114> for more information

error[E0658]: use of unstable library feature `inherent_str_constructors`
  --> src/macros/rustc.rs:46:4
   |
46 |             str::from_utf8(&output.stdout)
   |             ^^^^^^^^^^^^^^
   |
   = note: see issue #131114 <https://github.com/rust-lang/rust/issues/131114> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `tuwunel_macros` (lib) due to 2 previous errors
```

jevolk:
This is amazing! I'll look into the error and probably be able to add a CI step for nix now. Thank you both so much 🙏🏻

wkordalski:
> Can't seem to compile with `nix build github:wkordalski/tuwunel/nix` as it causes the following error in `rocksdb-tuwunel-changes`

Yes. It does not compile. Exactly: Tuwunel does not compile. Because it uses unstable API from standard library (`str::from_utf8`).
Probably the calls should be replaced with calls to other functions from stable Rust-standard-library API.
Or nightly Rust compiler should be used (depending on which Rust compiler version should Tuwunel depend on).

This PR fixes building RocksDB with Nix flake — now RocksDB is built correctly.
This PR does not fix compilation errors of Tuwunel itself.

coolGi69:
Ah alright.

PR #101, opened 2025-07-05
Enhanced docu for configuration

obioma:
Include that more than one config file can be named. 

Describe the relevance of ordering for setting the value.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -7,14 +7,15 @@ This chapter describes various ways to configure Tuwunel.
 Tuwunel uses a config file for the majority of the settings, but also supports
 setting individual config options via commandline.
 
-Please refer to the [example config
-file](./configuration/examples.md#example-configuration) for all of those
-settings.
+Please refer to the [example config file](./configuration/examples.md#example-configuration) for all of those settings.
 
 The config file to use can be specified on the commandline when running
-Tuwunel by specifying the `-c`, `--config` flag. Alternatively, you can use
-the environment variable `TUWUNEL_CONFIG` to specify the config file to used.
-Conduit's environment variables are supported for backwards compatibility.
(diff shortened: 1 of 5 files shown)


docs/configuration.md
```diff
@@ -4,17 +4,18 @@ This chapter describes various ways to configure Tuwunel.
 
 ## Basics
 
-Tuwunel uses a config file for the majority of the settings, but also supports
+Tuwunel uses a config file for the majority of the settings, but also supports
```
NinekoTheCat, line 7: Why is this changed by adding a single space character at the end?

NinekoTheCat (changes requested):
I only really left one comment about a strange change but that's about it

docs/configuration.md
obioma: Hi @NinekoTheCat . if you click on docs/configuration.md you will see it is more than this.

jevolk:
Thanks for making this contribution. I'm also curious about why trailing spaces are needed in addition to linebreaks, they're showing up for me as whitespace-errors. Is that required for highly-compatible markdown?

I'd like to merge this and you're under no obligation to do anything further; I'm just tempted to fix it while I do unless you have a reason why I shouldn't. Again thank you for taking the time to document this functionality 🙏🏻

obioma:
Hi @jevolk ,
Basically this was a kind of reflex when starting to edit the original file, which I have overseen later after @NinekoTheCat  pointed it out (sorry for that). I'm using RustRover, with hard wrap at 120 chars and soft wrap at 80 chars, and hides automatically link data, the (...) in the screen shots.

Thus, it looked like the upper screen shot, and when starting to enhance the description I edited it to look like in the lower screen, not thinking and no other reason than 'symmetry' in mind.

This means, you can change back or do whatever you want to to make it better.

Thanks for your hard work

image

Issue #99, opened 2025-07-04
Typing indicators not working


tmayoff:
SchildiChat and ElementX don't have typing indicators. 

I have a feeling this might also be related to #98 and #95. Continuwuity also seems to be working on [this](https://forgejo.ellis.link/continuwuation/continuwuity/pulls/842)  and I wanted to ensure it's tracked here as well

tmayoff:
I was planning on taking a stab at this as well. Starting by porting the changes made in the continuwity branch and going from there.

tmayoff:
This was merged but the issue didn't get closed, closing now

PR #107, opened 2025-07-11
Typing indicators

tmayoff:
closes: #99 

This is technically working ATM, but a little janky. 

- Element often makes 2 calls to `/sync` and the second clears the typing indicators. If you wait long enough it'll call `/sync` again and store the typing indicator. However if the other client does anything it'll be cleared immediately.

- This kind of avoids the todo_rooms APIs since for some reason it'll only make requests for 1 of the 2 test rooms I have.

### DIFF
--- a/src/api/client/sync/v3.rs
+++ b/src/api/client/sync/v3.rs
@@ -840,7 +840,7 @@ async fn load_joined_room(
 			let typings = services
 				.rooms
 				.typing
-				.typings_all(room_id, sender_user)
+				.typings_event_for_user(room_id, sender_user)
 				.await?;
 
 			Ok(vec![serde_json::from_str(&serde_json::to_string(&typings)?)?])
--- a/src/api/client/sync/v5.rs
+++ b/src/api/client/sync/v5.rs
@@ -18,6 +18,7 @@ use ruma::{
 	events::{
 		AnyRawAccountDataEvent, AnySyncEphemeralRoomEvent, StateEventType, TimelineEventType,
 		room::member::{MembershipState, RoomMemberEventContent},
+		typing::TypingEventContent,
 	},
 	serde::Raw,
 	uint,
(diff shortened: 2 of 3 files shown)


Cargo.toml
jevolk, line 1: I'm not certain but I would suspect at a glance the Ruma change is not necessary and there might be some other way to access that type implicitly.

tmayoff: You're right, rust-analyzer was broken when I started this and wasn't able to find the import

tmayoff:
Okay so tested this with element-web with siding sync enabled and it does work fine. Element-X on the other hand has some very weird behaviour.

[Screencast From 2025-07-11 22-40-54.webm](https://github.com/user-attachments/assets/f3c34676-a8ce-4e55-8ee3-dc80026b0e95). 

It's a touch hard to tell, but element-x keeps trying /sync it returns with typing info, then when you keep typing element-x calls /sync again and immediately clears it. tuwunel's logs show it's always returning the typing info

I'm not sure if it's due to the sync frequency (element-web seems to be going for a speed record for spamming the sync endpoint)

tmayoff:
Okay so I figured out the weirdness with Element-X, it's providing a conn_id and element-web isn't, which means the `roomsince` in element-x is always getting a new `roomsince` args and element-web is never getting a new `roomsince`. That at least explains the difference between the two behaviours but I'm once again not sure what the correct behaviour is it either never updates the known_room roomsince, or always does 

https://github.com/matrix-construct/tuwunel/blob/fd080914c2c4f9824bc911b83af9b7f12e0706de/src/api/client/sync/v5.rs#L375-L384

jevolk:
> once again not sure what the correct behaviour is it either never updates the known_room roomsince, or always does

Feel free to leave a comment on the MSC about this if it's not written out. It's not finalized yet last I checked...

tmayoff:
Okay looks like the main issue was the 30 second timeout, I tested with element-x in a simulator, the develop.element.io, slinding sync broke typing indicators I think, since it's still experimental I'm okay with skipping that.

This means that while typing clients will spam requests, which may not be desirable but maybe a smaller timeout would be better for these things

PR #110, opened 2025-07-21
Fixed nix cached shasum for rust-toolchain.toml

coolGi69:
This is a small fix to fix up the hash stored in the nix flake, making it possible to compile through nix again!\
(sending this PR in myself, as it has been broken for a few weeks, and I've hoped that someone else fixes this before me)

When updating the `rust-toolchain.toml` file, please remember to also update it's corresponding hash in the nix flake if possible.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -26,7 +26,7 @@
         file = ./rust-toolchain.toml;
 
         # See also `rust-toolchain.toml`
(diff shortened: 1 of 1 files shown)


jevolk:
Thank you for PR'ing this. What I really forgot to do here is add nix jobs to CI ensuring everything checks out.

PR #150, opened 2025-09-02
Push Suppression (Experimental Feature)

tototomate123:
This PR adds an **opt-in push suppression feature** to cut down on mobile notifications when a user is actively using another client. It's currently marked as **experimental** and is disabled by default.

#### **What's new**
- Added config flag:
  `suppress_push_when_active = true`
- Presence service now tracks the **timestamp of the last `/sync`** per user in memory.
- `/sync` calls update that timestamp automatically.
- Push sender checks:
  - user is `Online`
  - last active < ~65 s
  - last sync gap < ~32 s  
    => If all match, push delivery is **suppressed**; otherwise, notifications are sent as usual.

#### **Why the heuristic?**
Initially, I tried implementing this using presence alone, but that turned out to be too "sticky".  
Presence keeps users marked as `Online` for ~5 minutes (default config) after they stop syncing, so mobile notifications stayed suppressed much longer than intended.  

This update adds a simple heuristic based on the gap since the last `/sync` request. With this, push notifications resume after ~30 seconds of inactivity **without changing presence idle timers** or requiring extra presence updates.

#### **Notes**
- Fully **opt-in**, preserving legacy behavior by default.
- No schema or federation changes.
- If presence lookups fail, pushes are delivered as before.

#### **Next steps / ideas?**
- Make the thresholds configurable?
- Add metrics for suppressed vs delivered pushes?


jevolk:
Sorry about the delay in getting to this but I am definitely excited about it. Thank you so much for making this effort 🙏🏻

PR #152, opened 2025-09-03
Fix broken doctest in src/core/utils/debug.rs

scvalex:
Fix a broken documentation test.

I ran into this while [packaging 1.4.1 for NixOS](https://github.com/NixOS/nixpkgs/pull/439912) because we run tests by default.  I just patched the file there, so there's no rush to merge this.

I have not thought at all about what the code does—I just made it compile.


jevolk:
Ironically I was just mulling over doctests disappearing from the CI pipeline possibly around the first release, even though they were certainly intended. Thanks for this patch and I'm going to address that right now for sure 🙏🏻

scvalex:
Thank you!

PR #168, opened 2025-09-15
packaging: Move unit alias to install section

agx:
Otherwise there will be

```console
systemd[1]: /usr/lib/systemd/system/tuwunel.service:5: Unknown key 'Alias' in section [Unit], ignoring.
```

on startup. See `systemd.unit(5)`.


jevolk:
Thank you so much 🙏🏻 This and #173 will be going in on Monday/Tuesday.

PR #212, opened 2025-11-03
Fix shared TURN secret from `turn_secret_file` not being used

### LINKED ISSUE #211: `turn_secret_file` configuration option has no effect
When `turn_secret_file` is set, Tuwunel loads the secret from a file at the given path and stores it in its `globals` service:

https://github.com/matrix-construct/tuwunel/blob/9af15ecbba131c789495f1aa380e720412c23c55/src/service/globals/mod.rs#L25-L34

https://github.com/matrix-construct/tuwunel/blob/9af15ecbba131c789495f1aa380e720412c23c55/src/service/globals/mod.rs#L11-L16

This is implemented identically to registration tokens and works as intended. However, the stored value is never used, causing this configuration option to have no effect; when no other TURN credentials option is set (i.e. `turn_secret` or `turn_username/password`), the credentials advertised at `/_matrix/client/v3/voip/turnServer` are empty:

```json
{
  "username": "",
  "password": "",
  "uris": ["turn.example.org"],
  "ttl": 86400
}
```

Presumably, this needs to be fixed here:

https://github.com/matrix-construct/tuwunel/blob/9af15ecbba131c789495f1aa380e720412c23c55/src/api/client/voip.rs#L28-L30

where `&services.globals.turn_secret` should be used instead.

BVollmerhaus:
Fixes #211.

### DIFF
--- a/src/api/client/voip.rs
+++ b/src/api/client/voip.rs
@@ -25,7 +25,7 @@ pub(crate) async fn turn_server_route(
 		return Err!(Request(NotFound("Not Found")));
 	}
 
-	let turn_secret = &services.config.turn_secret;
+	let turn_secret = &services.globals.turn_secret;
 
 	let (username, password) = if !turn_secret.is_empty() {
 		let expiry = SecondsSinceUnixEpoch::from_system_time(


jevolk:
Thank you 🙏🏻

PR #221, opened 2025-11-14
Remove `allow_check_for_updates` from configuration files

### LINKED ISSUE #220: Remove Remains of Update Service
I've noticed the Tuwunel container reporting the following at every start:
> WARN tuwunel_core::config::check: Config parameter "allow_check_for_updates" is unknown to tuwunel, ignoring.

After some digging in the commit history I've found that this is most likely caused by the removal of the update service. The module itself is removed but not the environment variable which allowed to skip the update check. It can still found in various files, namely the compose files for Docker.

This commit removed the update service: edb245a2baec3e5aaaccdb99de7db1a8d4a903ea

I could create a pull request for removing the remains if my observations are considered valid.

Radiant-Xyz:
This PR fixes #220.

It removes a configuration variable used by the update module which was removed in edb245a2baec3e5aaaccdb99de7db1a8d4a903ea.

There are no changes to the actual code base so this should be not breaking anything.


jevolk:
Thank you so much 🙏. CI is down until monday but that's okay if you add `[ci skip]` otherwise it will be my pleasure to merge this normally when I get in. Thank you again 😊

Radiant-Xyz:
Hi there. Sorry for the delay. I've been out myself today.

`[ci skip]` has been added to the commit and the commit message was also a bit shortened by me to better suit the conventions. I hope this is alright.

The commit itself has not changed in content.

PR #234, opened 2025-12-05
Fix newly created LDAP users not added to auto-join-rooms

yefimg:
Users logged in via a LDAP server currently are not auto-joined into rooms listed in auto_join_rooms. This PR fixes that.


jevolk:
Thank you for contributing this 🙏🏻 . I am going to be on break for about a week so I can't review this until I get back. Please don't worry about any test flakes in CI, I'll have to deal with that. Again thank you!!

PR #259, opened 2026-01-16
Add Traefik reverse proxy documentation

Xerusion:
Updated the Traefik documentation with installation and configuration details.

### DIFF
--- a/docs/deploying/reverse-proxy-traefik.md
+++ b/docs/deploying/reverse-proxy-traefik.md
@@ -64,13 +64,24 @@ http:
                 passHostHeader: true
 ```
 ### Federation
+If you will use a .well-known file you can use traefik to redirect .well-known/matrix to tuwunel built-in .well-known file.
+
+replace the rule in either of the methods from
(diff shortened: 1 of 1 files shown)


jevolk:
Wow you are amazing!! 😍

Xerusion:
@jevolk should I remove the section about Encoded Character Filtering? They are now all true by default in version 3.6.7 and 2.11.35 or should I just keep the version warning?

jevolk:
I'm not an expert here but it caused a lot of grief only a few weeks ago... If it helps someone using a slightly older release/distro I would err on the side of caution and leave it if that makes sense.

Xerusion:
> I'm not an expert here but it caused a lot of grief only a few weeks ago... If it helps someone using a slightly older release/distro I would err on the side of caution and leave it if that makes sense.

Agreed.

This is done, if you are happy with the formatting and language.

Xerusion:
I think this is done, please review if I'm missing anything.

PR #264, opened 2026-01-20
 Presence aggregation + deferred push suppression

lhjt:
## Summary

- Aggregate per-device presence into a single user presence state with refresh-window logic. This means that multiple devices reporting different presences will no longer cause the user's presence to flip frequently between updates.
- Make presence timers abortable and resilient to stale timer fires.
- Add in deferred push notifications, as is in other platforms like Discord. The current implementation is in-memory only to avoid having to making a schema change whilst this feature is being evaluated. Example:
  1. push notification is suppressed
  2. user goes unavailable/offline (no longer online)
  3. any unread messages that arrived and were suppressed will be pushed to the user as push notifications

## Key changes

- Presence update pipeline extracted to `src/service/presence/pipeline.rs`
- New aggregation module `src/service/presence/aggregate.rs` and unit tests
- Presence timers now abort/reschedule safely; stale timers are ignored
- Deferred push suppression queue reintroduced (`src/service/pusher/suppressed.rs`) with enqueue/flush/clear hooks
- Presence inactivity now triggers suppressed push flush

## Behaviour notes

- Refresh-skip now requires `last_active_ago` to be present; missing values will write a new event instead of skipping
- Aggregation returns `Offline` immediately when all devices are pruned

## Tests

- `cargo test -p tuwunel_service presence`

### DIFF
--- a/src/service/presence/pipeline.rs
+++ b/src/service/presence/pipeline.rs
@@ -98,6 +98,7 @@ impl Service {
 		refresh_window_ms: Option<u64>,
 	) -> Result {
 		let now = tuwunel_core::utils::millis_since_unix_epoch();
+		// 1) Capture per-device presence snapshot for aggregation.
 		debug!(
 			?user_id,
 			?device_key,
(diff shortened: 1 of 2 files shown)


jevolk:
This is amazing work 😍

Thank you!!! 🙇🏻

PR #271, opened 2026-01-27
Don't log a password when a new user account is registered

bdfd9:
Introduced in 4150b917d826351c757fe35000f94114744097f4


jevolk:
Sorry about that 😭 Thanks for submitting the patch too

jevolk:
Please ignore the lints, I'll take it from here.

PR #275, opened 2026-01-30
Fix nix linkage issues

VlaDexa:
For some reason, after MSRV bump all the runpath stuff stopped writing itself to the binary. This adds mechanisms to put them back.

Would love for someone to fix that wihtout autoPatchelf for cleaner code.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,10 +1,11 @@
-(import
-  (
-    let lock = builtins.fromJSON (builtins.readFile ./flake.lock); in
-    fetchTarball {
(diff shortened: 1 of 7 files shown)


scvalex:
What does this PR fix?

Context: We're doing the 1.4.9.1->1.5.0 version bump in nixpkgs and I'm trying to figure out if we need these changes as well.  The server seems to work fine without them and I can't figure out what "runpath" refers to here.

VlaDexa:
This is related to the [MSRV bump](https://github.com/matrix-construct/tuwunel/commit/3398d1730e238710d7e2fff35b938d26a103a53a)

After switching versions the binary stopped linking itself to libs, so I had to link it by hand. 

You probably don't need to port this in any way because nixpkgs already been building with new rust versions

scvalex:
Got it.  Thank you!

PR #278, opened 2026-01-30
Docs: add authentication page

valentimarco:
As discussed in #274


jevolk:
Please note that I'm currently integrating this page with the SSO chapter in the docs that I'm working on. Unfortunately this has not been completed in time for the release which could not wait, but the docs are deployed separately anyway so this will be merged into main shortly.

PR #281, opened 2026-02-01
Rework oidc provider construction

### LINKED ISSUE #276: oAuth can't parse client's sso urls
2026-01-30T07:25:43.133788Z ERROR tuwunel_router::request: 500 Internal Server Error, method: GET, uri: https://mtx.domain.com/_matrix/client/v3/login/sso/redirect?redirectUrl=im.fluffychat%3A%2F%2Flogin
    at src/router/request.rs:116 on tuwunel:worker ThreadId(2)

  2026-01-30T07:25:43.134008Z ERROR tower_http::trace::on_failure: response failed, classification: Status code: 500 Internal Server Error, latency: 26 ms
    at /opt/rust/cargo/debian/x86_64-linux-gnu/registry/src/index.crates.io-1949cf8c6b5b557f/tower-http-0.6.8/src/trace/on_failure.rs:93 on tuwunel:worker ThreadId(2)

I have configured fully my OIDC SSO identity provider ,but still get this error 
config part:
    [[global.identity_provider]]

    brand = "VoidAuth"

    client_id = "ID"
    client_secret = "SECRET"

    issuer_url = "https://auth.domain.com/oidc"

    # The callback URL configured when registering the OAuth application with
    # the provider. Tuwunel's callback URL must be strictly formatted exactly
    # as instructed. The URL host must point directly at the matrix server and
    # use the following path:
    # `/_matrix/client/unstable/login/sso/callback/<client_id>` where
    # `<client_id>` is the same one configured for this provider above.
    #
    callback_url = "https://mtx.domain.com/_matrix/client/unstable/login/sso/callback/ID"

    default = true

    name = "VoidAuth"

    # Optional icon for the provider. The canonical providers have a default
    # icon based on the `brand` supplied above when this is not supplied. Note
    # that it uses an MXC url which is curious in the auth-media era and may
    # not be reliable.
    #
    #icon =

    # Optional list of scopes to authorize. An empty array does not impose any
    # restrictions from here, effectively defaulting to all scopes you
    # configured for the OAuth application at the provider. This setting
    # allows for restricting to a subset of those scopes for this instance.
    # Note the user can further restrict scopes during their authorization.
    #
    #scope = []

    # Optional list of userinfo claims which shape and restrict the way we
    # compute a Matrix UserId for new registrations. Reviewing Tuwunel's
    # documentation will be necessary for a complete description in detail. An
    # empty array imposes no restriction here, avoiding generated fallbacks as
    # much as possible. For simplicity we reserve a claim called "unique"
    # which can be listed alone to ensure *only* generated ID's are used for
    # registrations.
    #
    #userid_claims = []

    # Optional extra path components after the issuer_url leading to the
    # location of the `.well-known` directory used for discovery. This will be
    # empty for specification-compliant providers. We have supplied any known
    # values based on `brand` (e.g. `/login/oauth` for GitHub).
    #
    #base_path =
    discovery = true

It doesn't matter if i enable log= debug or log= info
Error is still the same

It just can't parse sso url:  (Tired  Element web,mobile(classic) and desktop versions, aswell as mobile fluffychat)

OIDC is configured correctly for sure,but it looks like i matrix can't parse this url to redirect it later to SSO

Oidc issuer is https://auth.domain.com/oidc/

ANd discovery is https://auth.domain.com/oidc/.well-known/openid-configuration

Used latest tagged release as well as main-branch 1 day ago released

Deployment is in container.

Jeidnx:
Rework the initialization code for oidc providers:

- Refactor the code to be more maintainable
- split github into its own init function since its a very special case
- Remove default values for {authorization,introspection,token} urls, since there is no standard default that could be used. All providers i know of support the discovery standard, which should be preferred instead of configuring these manually.
- Adds automatic construction of the callback url.
- Fixes #276

### DIFF
--- a/src/service/oauth/providers.rs
+++ b/src/service/oauth/providers.rs
@@ -212,6 +212,15 @@ async fn configure(&self, mut provider: Provider) -> Result<Provider> {
 			.map(|url| provider.token_url.replace(url));
 	}
 
+	if provider.callback_url.is_none()
+		&& let Some(server_url) = self.services.config.well_known.client.as_ref()
+	{
+		let callback_path =
+			format!("_matrix/client/unstable/login/sso/callback/{}", provider.client_id);
+
+		provider.callback_url = Some(server_url.join(&callback_path)?);
+	}
+
 	Ok(provider)
 }
 


src/service/oauth/providers.rs
```diff
@@ -235,25 +217,54 @@ pub async fn discover(&self, provider: &Provider) -> Result<JsonValue> {
...
+	}
+
+	if let Some(url) = &provider.discovery_url {
+		return Ok(url.to_owned());
+	}
+
+	let issuer = provider
+		.issuer_url
+		.as_ref()
+		.expect("Can we assert this exists at this point?");
```
Jeidnx: I think `issuer_url` gets asserted earlier in `configure` but im not really sure. Code is a bit confusing

src/service/oauth/providers.rs
```diff
@@ -235,25 +217,54 @@ pub async fn discover(&self, provider: &Provider) -> Result<JsonValue> {
...
-		.transpose()?;
-
-	let Some(url) = provider
-		.discovery_url
-		.clone()
-		.filter(|_| provider.discovery)
-		.or(default_url)
-	else {
+	if !provider.discovery {
 		return Err!(Config(
```
Jeidnx: It feels wrong to return an error in this case. Disabled discovery is not an error in my books. Maybe that should be checked before calling this function?

src/service/oauth/providers.rs
```diff
@@ -282,20 +293,21 @@ fn check_issuer(
...
+	response_map: &serde_json::Map<String, serde_json::Value>,
+	url_name: &str,
+) -> Result<Url> {
+	let Some(url_value) = response_map.get(url_name) else {
+		return Err(tuwunel_core::Error::Err(
+			format!(
+				"Error building oidc provider '{}': {} is missing from openid discovery response",
+				provider_id, url_name,
+			)
+			.into(),
```
Jeidnx: Also not sure if this is the correct way to return an error here. I've read the Err! macro description, but got even more confused.

x86pup: This can be shorted to:

```rs
		return Err!("Error building oidc provider '{provider_id}': {url_name} is missing from openid discovery response");
```

And importing `use tuwunel_core::Err`, if it's not already, for the `Err!` macro.

Jeidnx:
Sorry i may have been a bit harsh with the words. I was still in my Software Engineering mindset and honestly a little frustrated since it took me some time to get this software to compile properly. (I will probably open an issue or a PR for the problems i encountered)

I refactored a lot of the configuration code, in my opinion this should be way better to read and maintain.
Github seems to be a very special case, so it only made sense that it gets it own function. The other branded providers comply with the discovery spec.
The callback_url really shouldn't be configurable, i fixed that and indicated it in the docs.

If you have any questions about the code please don't hesitate to ask

Jeidnx:
Not sure why the CI fails on unchanged code. Runs fine on main: https://github.com/matrix-construct/tuwunel/actions/runs/21559255457/job/62120577532

x86pup:
I think this is some new issue with the typos tool we use for detecting typos, not the first time something like this has happened. We will handle it ourselves don't worry. Thank you for the contribution.

aazf:
@Jeidnx If possible, pls add username claim from OAuth provider.
some of provider return the username, not the preferred_username.

jevolk:
When testing GitHub I 404'ed after the initial redirect. At a glance it might be that the base_path is not being applied. I can have a look but I have to merge the other pending items to main branch first.

Jeidnx:
i rebased on main and made some changes to the github provider. I'm pretty sure this time the endpoints are the same as they where previously. Still couldn't find any documentation for them though. Btw did you find any or how do you know the endpoints?

I also decided to remove the `base_path`. There really isn't a reason to have this, because the only thing it would affect is the discovery url. If there is a provider where the discovery url is non standard, there is an explicit option to overwrite just that.

jevolk:
> I also decided to remove the base_path. There really isn't a reason to have this, because the only thing it would affect is the discovery url.

Unfortunately we can't remove this because it would break released code. The only way to effectively deprecate it at this point is to still apply the value somewhere expected when still used.

jevolk:
Hypothetically, would adjusting the path literals from the original code leading to `Url::join()` like https://github.com/matrix-construct/tuwunel/commit/f6c3c31ded4a9889527f44e27b7889ddd3727d49 fix the specific `issuer_url` issue?

I like the automatic callback_url from this pr as well, that's a great idea.

Jeidnx:
Not quite, for `Url::join()` it matters both if there is a  slash at the end of the `Url`, and if there is one at the start of the input. I've assembled a minimal patch that according to my testing fixes the discovery issue. If that is what you want, feel free to apply that and close this PR:
[fix_discovery.patch](https://github.com/user-attachments/files/25160466/fix_discovery.patch)

PR #299, opened 2026-02-10
rm version field from turn docker compose example

Lamby777:
It was giving errors because the docs put 3 as a number and not a string.
When fixed, it still gave me warnings saying that field is obsolete and saying to remove it, so yea


jevolk:
Thank you for fixing this! Don't worry about the CI, we got this.

PR #342, opened 2026-02-26
feat: implement OIDC server for next-gen auth (MSC2965/2964/2966/2967)

### LINKED ISSUE #13: Implement MSC2964
Implement [MSC2964](https://github.com/matrix-org/matrix-spec-proposals/pull/2964) titled: Usage of OAuth 2.0 authorization code grant and refresh token grant

### LINKED ISSUE #14: Implement MSC2965
Implement [MSC2965](https://github.com/matrix-org/matrix-spec-proposals/pull/2965) titled OAuth 2.0 Authorization Server Metadata discovery

### LINKED ISSUE #15: Implement MSC2966
Implement [MSC2966](https://github.com/matrix-org/matrix-spec-proposals/pull/2966) titled Usage of OAuth 2.0 Dynamic Client Registration in Matrix

### LINKED ISSUE #16: Implement MSC2967
Implement [MSC2967](https://github.com/matrix-org/matrix-spec-proposals/pull/2967) titled API scopes

### LINKED ISSUE #232: Error: "Your homeserver needs to be upgraded to support matrix authentication service" on registration attempt in Element X
I run tuwunel on my VPS, and creating users via admin commands works as intended. But when I try to register a new user via Element X app using my VPS as the provider, I'm getting an error:
```
Your homeserver needs to be upgraded to support matrix authentication service
````
I have `allow_registration = true` and `registration_token` set in my config.

Does Tuwunel support MAS? I've read the [MAS docs](https://element-hq.github.io/matrix-authentication-service/index.html), but it only mentions Synapse as supported homeserver.

### PR DESCRIPTION
**Pull request by lytedev**
Implements a built-in OIDC authorization server that allows Matrix clients like Element X to authenticate via the next-gen auth flow (MSC2964). User authentication is delegated to upstream identity providers (e.g. Kanidm) through the existing SSO/OAuth client flow.

## Endpoints
- auth_issuer + auth_metadata discovery (stable v1 + unstable MSC2965)
- OpenID Connect discovery (/.well-known/openid-configuration)
- Dynamic Client Registration (MSC2966)
- Authorization + token + revocation + JWKS + userinfo
- SSO bridge: authorize → SSO redirect → complete → code → token

## Features
- ES256 (P-256) JWT signing with persistent key material
- PKCE (S256) support with RFC 7636 verifier validation (43-128 chars, restricted charset)
- Authorization code grant with refresh tokens
- All OIDC state persisted in RocksDB (signing keys, client registrations, auth codes, pending auth requests)
- Device ID extraction from MSC2967 scopes

## Spec compliance fixes
- OAuth error responses use RFC 6749 §5.2 format (`{"error": "...", "error_description": "..."}`) on token, registration, and revocation endpoints instead of Matrix `{"errcode": "M_..."}` format
- PKCE code_verifier validation per RFC 7636 §4.1 (length and charset checks)
- Scope token matching uses exact whitespace-delimited comparison per RFC 6749 §3.3 instead of substring matching
- Typed `ProviderMetadata` struct for the discovery document with documented fields per RFC 8414 / OpenID Connect Discovery 1.0
- DCR request/response includes `policy_uri`, `tos_uri`, `software_id`, `software_version` per RFC 7591

Refs: #246, #266

### CODE REVIEW

**Comment by Ludea** [comment]
There is https://github.com/element-hq/matrix-authentication-service/tree/main/crates/cli which start MAS

**Comment by lytedev** [comment]
> There is https://github.com/element-hq/matrix-authentication-service/tree/main/crates/cli which start MAS

Looks like that code is under an incompatible software license. I'm a big AGPL fan, but we can't use it here as far as I'm aware =(

**Comment by chbgdn** [comment]
Working great. Added some fixes for Element X (it doesn't request openid scope, which is not mandatory) and implemented real client names (switched from hardcoded "OIDC Client" to the registered name with a fallback).

```
diff --git a/src/api/client/oidc.rs b/src/api/client/oidc.rs
index 73eaac2a..aac7881c 100644
--- a/src/api/client/oidc.rs
+++ b/src/api/client/oidc.rs
@@ -76,8 +76,6 @@ pub(crate) async fn authorize_route(State(services): State<crate::State>, reques
 
 	oidc.validate_redirect_uri(&params.client_id, &params.redirect_uri).await?;
 
-	if !scope_contains_token(&params.scope, "openid") { return Err!(Request(InvalidParam("openid scope is required"))); }
-
 	let req_id = utils::random_string(OIDC_REQ_ID_LENGTH);
 	let now = SystemTime::now();
 
@@ -147,8 +145,10 @@ async fn token_authorization_code(services: &tuwunel_service::Services, body: &T
 	let (access_token, expires_in) = services.users.generate_access_token(true);
 	let refresh_token = generate_refresh_token();
 
+	let client_name = oidc.get_client(&session.client_id).await.ok().and_then(|c| c.client_name).unwrap_or_else(|| "OIDC Client".to_owned());
+
 	let device_id: Option<OwnedDeviceId> = extract_device_id(&session.scope).map(OwnedDeviceId::from);
-	let device_id = services.users.create_device(user_id, device_id.as_deref(), (Some(&access_token), expires_in), Some(&refresh_token), Some("OIDC Client"), None).await?;
+	let device_id = services.users.create_device(user_id, device_id.as_deref(), (Some(&access_token), expires_in), Some(&refresh_token), Some(&client_name), None).await?;
 
 	info!("{user_id} logged in via OIDC (device {device_id})");
```

**Comment by jevolk** [comment]
Thank you both! I'm excited to have a go at this soon.

**Comment by siennathesane** [comment]
Tested this with Element X iOS against our production tuwunel instance (Hydra as the upstream IdP) — works great! Two small changes we needed:

  1. Remove hard `openid` scope requirement in `authorize_route`

Same fix @chbgdn posted above. Element X doesn't request the `openid` scope, which is optional per OIDC spec. The `id_token` generation in `token_authorization_code` already handles this correctly (only mints one when `openid` is in scope).

```
  -     if !scope_contains_token(&params.scope, "openid") { return Err!(Request(InvalidParam("openid scope is required"))); }
  -     let req_id = utils::random_string(OIDC_REQ_ID_LENGTH);
```

  2. Use DCR-registered client name as device display name

Without this, every OIDC session shows up as "OIDC Client" in the device list. With it, you get "Element X" (or whatever the client registered as via DCR).

```
        let device_id: Option<OwnedDeviceId> = extract_device_id(&session.scope).map(OwnedDeviceId::from);
  -     let device_id = services.users.create_device(user_id, device_id.as_deref(), (Some(&access_token), expires_in), Some(&refresh_token), Some("OIDC Client"), None).await?;
  +     let client_name = oidc.get_client(client_id).await.ok().and_then(|c| c.client_name);
  +     let device_display_name = client_name.as_deref().unwrap_or("OIDC Client");
  +     let device_id = services.users.create_device(user_id, device_id.as_deref(), (Some(&access_token), expires_in), Some(device_display_name), None).await?;
```

Thanks @lytedev and @chbgdn for the great work on this — was very straightforward to deploy.

//cc @lonnithelost

**Comment by DonPrus** [comment]
Tested this locally on top of the PR, and these two changes were enough for me as well.

I brought up a local isolated OIDC-capable test setup and verified both paths against the real router/handlers:
- `/authorize` now succeeds without `openid` in the requested scope
- `/token` uses the DCR-registered `client_name` as the created device display name
- no `id_token` is returned when `openid` is not in scope

Path is the same as @siennathesane mentioned

**Comment by shaba** [comment]
Fixes OIDC loopback redirects for native clients like Fractal. Previously
  validate_redirect_uri() required an exact string match, so callbacks like http://127.0.0.1:<random-port>/ failed with M_INVALID_PARAM:
  redirect_uri not registered for this client.

  Now loopback HTTP redirect URIs on 127.0.0.1 / ::1 are matched ignoring only the port, while still requiring the rest of the URI to  match. Non-loopback redirect URIs still use exact matching.
That makes the OIDC flow work for Fractal and other native apps using RFC 8252-style loopback callbacks, without loosening redirect  validation for non-loopback URIs.
```

diff --git a/src/service/oauth/oidc_server.rs b/src/service/oauth/oidc_server.rs
index 7faac23eb..758fd0dfb 100644
--- a/src/service/oauth/oidc_server.rs
+++ b/src/service/oauth/oidc_server.rs
@@ -1,4 +1,4 @@
-use std::{sync::Arc, time::{Duration, SystemTime}};
+use std::{net::IpAddr, sync::Arc, time::{Duration, SystemTime}};
 
 use base64::{Engine as _, engine::general_purpose::URL_SAFE_NO_PAD as b64};
 use ring::{rand::SystemRandom, signature::{self, EcdsaKeyPair, KeyPair}};
@@ -141,7 +141,11 @@ pub async fn get_client(&self, client_id: &str) -> Result<OidcClientRegistration
 
 	pub async fn validate_redirect_uri(&self, client_id: &str, redirect_uri: &str) -> Result {
 		let client = self.get_client(client_id).await?;
-		if client.redirect_uris.iter().any(|uri| uri == redirect_uri) { Ok(()) } else { Err!(Request(InvalidParam("redirect_uri not registered for this client"))) }
+		if client.redirect_uris.iter().any(|uri| redirect_uri_matches(uri, redirect_uri)) {
+			Ok(())
+		} else {
+			Err!(Request(InvalidParam("redirect_uri not registered for this client")))
+		}
 	}
 
 	pub fn store_auth_request(&self, req_id: &str, request: &OidcAuthRequest) { self.db.oidcreqid_authrequest.raw_put(req_id, Cbor(request)); }
@@ -215,3 +219,25 @@ pub fn sign_id_token(&self, claims: &IdTokenClaims) -> Result<String> {
 
 	#[must_use] pub fn auth_request_lifetime() -> Duration { AUTH_REQUEST_LIFETIME }
 }
+
+fn redirect_uri_matches(registered: &str, requested: &str) -> bool {
+	if registered == requested {
+		return true;
+	}
+
+	match (url::Url::parse(registered), url::Url::parse(requested)) {
+		(Ok(registered), Ok(requested)) if is_loopback_redirect_uri(&registered) && is_loopback_redirect_uri(&requested) => {
+			registered.scheme() == requested.scheme()
+				&& registered.host_str() == requested.host_str()
+				&& registered.path() == requested.path()
+				&& registered.query() == requested.query()
+				&& registered.fragment() == requested.fragment()
+		},
+		_ => false,
+	}
+}
+
+fn is_loopback_redirect_uri(uri: &url::Url) -> bool {
+	uri.scheme() == "http"
+		&& matches!(uri.host_str().and_then(|host| host.parse::<IpAddr>().ok()), Some(ip) if ip.is_loopback())
```

lytedev:
Thank you all for these fine additions! For the sake of simplicity and for updating my own deployment. I've gone ahead and included them in a new commit on this PR.

jevolk:
Would it be too much to ask if you could rebase on the latest main branch so we could get CI rolling here?

lytedev:
Not at all. Will do so later today!

lytedev:
Rebased!

chbgdn:
@lytedev Since account/device management panel aren't implemented yet, the stub HTML page is pretty useless. Instead of returning a "not implemented" message, we could just redirect to the IdP's `issuer_url` where users can actually manage their account.

Patch:

```diff
diff --git a/src/api/client/oidc.rs b/src/api/client/oidc.rs
--- a/src/api/client/oidc.rs
+++ b/src/api/client/oidc.rs
@@ -452,11 +452,30 @@ pub(crate) async fn userinfo_route(
 	})))
 }
 
-pub(crate) async fn account_route() -> impl IntoResponse {
-	axum::response::Html(
-		"<html><body><h1>Account Management</h1><p>Account management is not yet implemented. \
-		 Please use your identity provider to manage your account.</p></body></html>",
-	)
+pub(crate) async fn account_route(
+	State(services): State<crate::State>,
+) -> Result<impl IntoResponse> {
+	// Redirect to the identity provider's panel where users can manage
+	// their account, sessions, devices, and profile.
+	let idp = services
+		.config
+		.identity_provider
+		.values()
+		.find(|idp| idp.default)
+		.or_else(|| services.config.identity_provider.values().next())
+		.ok_or_else(|| err!(Config("identity_provider", "No identity provider configured")))?;
+
+	let panel_url = idp
+		.issuer_url
+		.as_ref()
+		.ok_or_else(|| {
+			err!(Config(
+				"issuer_url",
+				"issuer_url is required for account management redirect"
+			))
+		})?;
+
+	Ok(Redirect::temporary(panel_url.as_str()))
 }

lytedev:
Massive. Thank you, @jevolk! :pray:

jevolk:
It wouldn't have been possible without you! And I also owe huge thanks to everyone who tested and patched during the run-up here to really drive this home!

siennathesane:
Thank you for this!!

Noah-Silvera:
Thank you @jevolk !

CEbbinghaus:
@lytedev @jevolk I appreciate your work and having completed this monumental feature, I would however like to raise an inherent flaw in this and every other implementation. In particular an exploit which arises when backing any [RFC7591](https://datatracker.ietf.org/doc/html/rfc7591) (DCR) implementation on a database (especially bad with an implementation as in this PR).

As every single DCR registration creates a new database entry and we don't limit any of the field lengths, it allows for creating megabytes of data in the database with every request, and given this is an unauthenticated endpoint that opens every single instance to (D)DOS by writing data into the db until the host runs out of storage.

Especially bad is that this implementation creates a new database entry for **every. single. request.** malicious or not. The matrix.org implementation caches based on the details provided by the client and reuses client_id for any known clients, so at the very least under "normal" circumstances there will only ever be a small handful of records in the DCR table.

But here, every legitimate authentication request will write up to 1kb of data, every time anyone logs in. And illegitimate registrations are able to write unbounded amounts of data into the database unauthenticated.

So while I 100% want to see some limitations being put in place (such as limiting the client_name to X (configurable) number of characters, and same for any other field) I would much prefer a solution that circumvents having to write data to the database completely.

I have a design we at kanidm have come up with to allow for entirely stateless DCR, which solves all of the above, happy to share in a different issue if requested

jevolk:
> @lytedev @jevolk I appreciate your work and having completed this monumental feature, I would however like to raise an inherent flaw in this and every other implementation. In particular an exploit which arises when backing any [RFC7591](https://datatracker.ietf.org/doc/html/rfc7591) (DCR) implementation on a database (especially bad with an implementation as in this PR).
> 
> As every single DCR registration creates a new database entry and we don't limit any of the field lengths, it allows for creating megabytes of data in the database with every request, and given this is an unauthenticated endpoint that opens every single instance (D)DOS by writing data to db until the host runs out of storage.
> 
> Especially bad is this implementation creates a new database entry for **every. single. request.** malicious or not. The matrix.org implementation caches based on the details provided by the client and reuses client_id for any known clients, so at the very least under "normal" circumstances there will only ever be a small handful of records in the DCR table.
> 
> But here, every legitimate authentication request will write up to 1kb of data, every time anyone logs in. And illegitimate registrations are able to write unbounded amounts of data into the database unauthenticated.
> 
> So while I 100% want to see some limitations being put in place (such as limiting the client_name to X (configurable) number of characters, and same for any other field) I would much prefer a solution that circumvents having to write data to the database completely.
> 
> I have a design we at kanidm have come up with to allow for entirely stateless DCR, which solves all of the above, happy to share in a different issue if requested

fix staged here: https://github.com/matrix-construct/tuwunel/commit/7fcd045f385d1b336a89522ea269662c61231266

CEbbinghaus:
@jevolk I love the prompt response, and the fix looks to be addressing the issue of filling up storage space, which addresses the core vulnerability behind this implementation.

I would however like to see the ability to limit size of registration, ideally measured as bytes of initial request as very few clients will exceed the 1kb limit, this allows upwards of 65536 clients to be registered within the 64Mb cache while denying large malicious registrations. As currently a single 64Mb registration (trivial to generate) would chew up the entirety of the cache and force all previously registered clients to be evicted.

Issue #263, opened 2026-01-20
Implement MSC2246


VlaDexa:
MSC https://github.com/matrix-org/matrix-spec-proposals/pull/2246
> Sending media to Matrix currently requires that clients first upload the media to the content repository and then send the event. This is a problem for some use cases, such as bridges that want to preserve message order, as reuploading a large file would block all messages.

PR #347, opened 2026-02-28
add MSC2246 support

donjuanplatinum:
#263 
Following the implementation of [https://github.com/matrix-org/synapse/pull/15503](https://github.com/matrix-org/synapse/pull/15503).


jevolk:
Thank you so much for putting this together 🤩. I will be on this asap to see if we can cut it into the pending release 🙏

PR #354, opened 2026-03-05
:memo:(docs) Add testmatrix example to troubleshooting, podman quadlet examples, and OIDC Keycloak example

rexbron:
Add a section to the docs on MatrixRTC testing tools. 

I have added it onto the Troubleshooting section.


rexbron:
I've also added podman quadlet examples and keycloak ODIC examples that will rely on the SSO changes in >=1.5.1.

Issue #355, opened 2026-03-06
[Feature request] Add support for systemd socket activation


ventureoo:
systemd socket activation - is the way to passthrough a socket that has been created in advance by systemd into a service without the need for an application to maintain its life cycle [1]. Most web servers like nginx or apache work well with this mechanism [2], but unfortunately tuwunel cannot for S2S connections.

This is particularly unpleasant in containerized environment, such as podman, which allows you to use systemd socket activation for containers, so that it would simultaneously preserve performance and work within a internal network without losing the ability to see the real IP of clients [3].

In general, this will only require tuwunel to check if `LISTEN_FDS` variable is present at launch and if it's, duplicate socket using this descriptor instead of creating a new one. It also possible to use listenfd crate for this purpose [4].

It would be nice to see that tuwunel can support this to make Matrix federation working.

Relevant links:
[1] - https://0pointer.de/blog/projects/socket-activation.html
[2] - https://systemd.io/DAEMON_SOCKET_ACTIVATION
[3] - https://github.com/containers/podman/blob/main/docs/tutorials/socket_activation.md
[4] - https://github.com/mitsuhiko/listenfd

PR #360, opened 2026-03-08
Add support for systemd socket activation

ventureoo:
I decided to use sd_notify crate because it is already used in the codebase. Also note that this does not take into account activation of Unix sockets via systemd, as I am not sure how best to implement this given that tuwunel implies using only one Unix socket, whereas systemd allows multiple. But since the main use-case is TCP sockets, it works just fine for me.

Closes: https://github.com/matrix-construct/tuwunel/issues/355

P. S. I’m a bit new to Rust, so feel free to say if this code is really bad.

### DIFF
--- a/src/router/serve.rs
+++ b/src/router/serve.rs
@@ -3,8 +3,6 @@ mod plain;
 mod tls;
 mod unix;
 
-#[cfg(all(feature = "systemd", target_os = "linux"))]
-use std::os::fd::FromRawFd;
 use std::{
 	net::{SocketAddr, TcpListener},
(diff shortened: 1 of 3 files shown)


ventureoo:
v2: I think I’ve fixed compilation on systemd-free targets and also squash the formatting fixes made by @jevolk (thanks for that!)

jevolk:
> v2: I think I’ve fixed compilation on systemd-free targets and also squash the formatting fixes made by @jevolk (thanks for that!)

Yes that's what I was going to ask for thank you!

PR #362, opened 2026-03-09
Add support for S3 storage provider

### LINKED ISSUE #21: Adding S3 client | port of #93
original: https://github.com/girlbossceo/conduwuit/issues/93
## TLDR
Tuwunel currently stores large media files by itself, we need to add a way to be able to use an S3 compatible store to store large media files in. It could potentially lessen the load in environments that support it.

S3 compatible stores support returning links for files which we could use to delegate the handling of files when it is supported.

### Tasks after this
Implement migration from static store to S3 compat store.
Implement migration from S3 to static store.

exodrifter:
When defined in the configuration, S3 object storage is used instead of the local filesystem for storing media. It is implemented by downloading the media from the bucket and then sending it to the user. This is kind of what is done today if you use s3fs with Tuwunel, except it doesn't do any additional work that s3fs can do like caching.

In the future, it would be nice to send the user a redirect as suggested by MSC3860 instead of relaying the data. However, ruma doesn't appear to support this kind of response so I've left that out for now.

A few changes not directly related to S3:
- Renamed `FileMeta` to `Media`, which feels more accurate to me.
- Admin file info command now returns metadata directly, which allows us to make `Media.content` non-nullable.
- Disambiguates the difference between media file name and path functions.
- ~~Fixed bounds being treated incorrectly for `!admin media delete-past-remote-media`~~ Moved to #375

Closes #21.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -679,7 +679,10 @@ version = "0.4.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c673075a2e0e5f4a1dde27ce9dee1ea4558c7ffe648f576438a20ca1d2acc4b0"
 dependencies = [
+ "iana-time-zone",
  "num-traits",
+ "serde",
+ "windows-link",
 ]
 
 [[package]]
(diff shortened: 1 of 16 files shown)


exodrifter:
FYI I plan on reverting 9c32adc and opening a new PR to address the remote media deletion command which integrates some of the feedback I received in the Matrix chat concerning the changes (renaming the arguments so people don't accidentally use it wrong with the new behaviour). However, I'm currently preoccupied with the Game Developers Conference so I will not be doing that until next weekend at the earliest.

jevolk:
Thank you for the amazing work on this issue which so many people will love. I did not quite get to this myself last week and I'm currently off for several days. Expect the full red carpet for this as soon as I return 😍

exodrifter:
You're welcome @jevolk, I'm very happy that the work is appreciated. :heart: Please let me know if there's anything else I can do!

I've just moved the remote media command changes to #375.

PR #386, opened 2026-03-21
fix(sync): stop prior long-poll on reconnection

chocycat:
When a client disconnects (browser refresh or network drop) during a sliding sync long-poll, the server continues to hold the connection mutex until the full timeout expires. Any new request from the same client on the same `conn_id` blocks on acquiring this lock, causing the user to wait for the previous request's timeout before the sync resumes.

This adds a `Notify` to each connection entry, so when a new request arrives for the same `(user, device, conn_id)`, it notifies the previous handler, which then exits the long-poll loop and releases the mutex. The new request proceeds immediately.

# Reproduction
1. Start a sliding sync long-poll with timeout=30000
2. While it's active, make another request with the same conn_id
3. Before the fix: the second request blocks for ~30s
4. After the fix: the second request proceeds immediately

Tested on Tuwunel v1.5.1 with a live server.

### DIFF
--- a/src/api/client/sync/v5.rs
+++ b/src/api/client/sync/v5.rs
@@ -101,7 +101,7 @@ pub(crate) async fn sync_events_v5_route(
 		.unwrap_or(0);
 
 	let conn_key = into_connection_key(sender_user, sender_device, request.conn_id.as_deref());
-	let conn_val = services
+	let (conn_val, superseded) = services
 		.sync
 		.load_or_init_connection(&conn_key)
 		.await;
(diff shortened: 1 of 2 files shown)


jevolk:
Thank you so much for finding this issue. This is actually a really delicate matter given some of the specific quirks discovered in the SDK and sliding sync. I'm going to need to review this thoroughly.

jevolk:
Apologies this didn't make it into 1.6.0. I'll be slightly tweaking (and abstracting) the precise way disconnections interrupt the request in progress after merging this. Thank you for your patience 🙏🏻

PR #387, opened 2026-03-21
added user_id in user registration notice

Lama-Thematique:
Simple and short PR for convernience. 

I often have to run command in the admin rooms for new users using their  ID, It would be convenient to have their name directly in the notice sent to the admin room. Currently I have to repeat `!admin users list` and try to find where is the new user in that list each time.

### DIFF
--- a/src/api/client/register.rs
+++ b/src/api/client/register.rs
@@ -381,7 +381,7 @@ pub(crate) async fn register_route(
 	if body.appservice_info.is_none() && (!is_guest || services.config.log_guest_registrations) {
 		let mut notice = String::from(if is_guest { "New guest user" } else { "New user" });
 
-		write!(notice, " registered on this server from IP {client}")?;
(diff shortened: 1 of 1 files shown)


Lama-Thematique:
Also i'm not sure why "Lint / Clippy / clippy test stable default debian testing-slim x86_64-v1-linux-gnu" is failling. Is there any additional fix needed from me for that ?

jevolk:
> Also i'm not sure why "Lint / Clippy / clippy test stable default debian testing-slim x86_64-v1-linux-gnu" is failling. Is there any additional fix needed from me for that ?

This flake was caused by a docker issue in our CI, sorry about that!

Just to reiterate your question in the chat: small QOL PR's like this are very much welcome and appreciated; feel free to submit as many as you'd like, one-liners very much accepted. Thank you for this!

Issue #314, opened 2026-02-15
SSO auth cannot be used to logout of sessions


exodrifter:
If you have only SSO authentication, you cannot logout of sessions. This is because if you try to log out of a session the user will be asked to do password authentication, but the account doesn't have a password. I believe this may happen because the only advertised authentication flows are password and jwt:

https://github.com/matrix-construct/tuwunel/blob/6c91aa1ddcfc2c25202dc9a1700ee7d11305fd6a/src/api/router/auth/uiaa.rs#L19-L22

My expectation is that if the server uses single sign on, when you try to log out of a session (or any other action which requires re-authenticating), these actions would do SSO authentication.

Related:
- https://github.com/element-hq/element-web/issues/20733

hatomist:
should be fixed by https://github.com/matrix-construct/tuwunel/pull/313; but old registrations on SSO only might need migration, as the origin for those is already set to "password" instead of "sso"

though it is a partial fix I believe, with correct origin tuwunel would fall through "no auth required" for removing the sessions; instead of requiring an SSO login. which is potentially a security risk, but I'm not that familiar with how reference clients do that. assume it should ask for SSO login as shown on the screenshot in the related link, and server should check sso response instead of no auth

alaviss:
For some more context, this is Tuwunel missing an implementation of [MSC2454](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2454-ui-interactive-auth-for-sso.md).

PR #388, opened 2026-03-22
fix: return m.change_password capability based on login_with_password config

chbgdn:
Sets the `m.change_password` capability based on the `login_with_password` config option. By default, the capability returns `enabled: true`, but now it will correctly return `enabled: false` when password login is disabled (e.g., SSO-only setups).

This allows clients to correctly determine whether password changes are supported.

### DIFF
--- a/src/api/client/capabilities.rs
+++ b/src/api/client/capabilities.rs
@@ -6,8 +6,9 @@ use ruma::{
 	api::client::discovery::{
 		get_capabilities,
 		get_capabilities::v3::{
-			Capabilities, GetLoginTokenCapability, ProfileFieldsCapability, RoomVersionStability,
-			RoomVersionsCapability, ThirdPartyIdChangesCapability,
+			Capabilities, ChangePasswordCapability, GetLoginTokenCapability,
+			ProfileFieldsCapability, RoomVersionStability, RoomVersionsCapability,
+			ThirdPartyIdChangesCapability,
 		},
 	},
 };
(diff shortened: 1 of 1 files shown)


jevolk:
Thank you for this but I don't think we can merge it until #314 is closed as SSO users must be able to set a password right now to use UIA-requiring features such as session removal.

chbgdn:
@jevolk I actually have a working branch that implements SSO UIA, which would allow SSO users to use UIA-requiring features. The code isn't perfect, but it's tested and working.

Would you like me to open it as a draft PR so you can see the approach?

jevolk:
> @jevolk I actually have a working branch that implements SSO UIA, which would allow SSO users to use UIA-requiring features. The code isn't perfect, but it's tested and working.
> 
> Would you like me to open it as a draft PR so you can see the approach?

Wow ABSOLUTELY! 🙏 I will be happy to pick up where you left off. This is a high priority issue going into the next dev cycle.

chbgdn:
> Wow ABSOLUTELY! 🙏 I will be happy to pick up where you left off. This is a high priority issue going into the next dev cycle.

I've opened this as a draft PR (#389).

Just to set expectations - this is more of a prototype/proof of concept. I'm not sure I have the skills/experience to polish it to merge-ready state, but hopefully the implementation can be useful as a reference or starting point.

Feel free to take ideas/code from it if helpful, or I can try to improve it with your guidance.

PR #392, opened 2026-03-23
Change default IP lookup strategy to IPv6 with IPv4 as fallback

jameskimmel:


x86pup:
What's the rationale for this default change? This can be changed in your config if you find it necessary.

The issue is this is going to make federation less performative because the majority of the Internet still is IPv4, and there are still IPv6 deployments out there that are less efficient or outright broken.

jameskimmel:
> What's the rationale for this default change? 

Short: Happy Eyeballs

Long: 
IMHO if you use IPv6 only as a fallback, you won't notice bugs in your IPv6 implementation. 
Maybe even some strange edge cases could be prevented. For example, I recently noticed that the Windows Element Desktop client (and just that, not the Element Client on macOS nor https://app.element.io/) had an issues with changing the password when there is no SplitDNS or Hairpin NAT set for tuwunel.  
Granted, since I use a reverse proxy, that SplitDNS would be needed either way, but if I would not use a proxy, IPv6 would work out of the box (because it uses something that is universally routable without overrides, like for example 2000:1234:1234:1234::1, while IPv4 (80.80.80.80) won't work work without any overrides or HairpinNAT.

I lack right now the imagination and technical knowledge to proof this, but it would not surprise me, if this would also be beneficial for calls. 

> there are still IPv6 deployments out there that are less efficient or outright broken.

which you could argue, then can use IPv4 as fallback.

x86pup:
I still don't see how this is a valid explanation for changing the default for everyone. It moreso reads like an issue on your end only, which is why the config option for this exists.

I'm personally an advocate for more IPv6 usage, but this is not the way to do it. Tuwunel, and previously conduwuit, is designed to be high performance and low latency. This is going to make a noticeable impact in negative federation performance for the majority of our userbase. The default is perfectly acceptable and most compatible here.

jevolk:
Matrix is quite possibly the heaviest commodity application for DNS load in modern existence. I can't think of another application installed en masse placing this kind of load on DNS. I used to work with IRC servers that had special requirements for rapid PTR queries- but only one party runs that, everyone else just runs a client.

We are constantly dealing with DNS issues while supporting diverse deployments with different service levels. We just have zero headroom for even a single wasted query at this point just to make federation work.

jameskimmel:
Ohh you misunderstood, I am not having issues. 
I did this because I expected the oppostite of 
> This is going to make a noticeable impact in negative federation performance for the majority of our userbase

PR #406, opened 2026-04-05
Route ephemeral events to appservices with receive_ephemeral

### LINKED ISSUE #382: Appservices with receive_ephemeral should receive EDUs via namespaces
Per MSC2409, an AS with `receive_ephemeral: true` must receive these events if the room or user matches its registered namespaces, regardless of membership.
Commit https://github.com/matrix-construct/tuwunel/commit/3675c941f8fe4b92b5426f24a667e0b03cabd9da attempted to fix this but caused server hangs/deadlocks and was reverted in https://github.com/matrix-construct/tuwunel/commit/f54a62dda0e989eda9a28f09121a142f692638c2.

References:
https://github.com/mautrix/signal/issues/580#issuecomment-2614104734

chbgdn:
Appservices with `receive_ephemeral: true` now receive `m.typing` and `m.receipt` events for rooms matching their registered namespaces.

Fixes #382.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -4,10 +4,4 @@ The release pipeline `Main` (main.yml) and its subroutines defined in the other 
 description for the underlying self-hosted build system in  `/docker`. In other words, this is a sort of
 terminal, a "thin-client" with a display and a keyboard for our docker mainframe. We minimize
 vendor-lockin and duplication with other services by limiting everything here to only what is
-essential for driving the docker builder.
-
-Though we slightly relax the above by specifying details of the actual CI pipeline, the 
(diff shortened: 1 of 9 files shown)


gymnae:
Would this bug fix help propagate read receipts? For example, if a message has been read in a room with a double puppet active from a mautrix bridge, would the corresponding messenger "learn" that the message has been read?

jevolk:
> Would this bug fix help propagate read receipts? For example, if a message has been read in a room with a double puppet active from a mautrix bridge, would the corresponding messenger "learn" that the message has been read?

I believe so.

@chbgdn Thanks for rebasing this on the main branch. I'll see what I can do about this right now. The challenge has been the lack of appservice testing infrastructure.

PR #410, opened 2026-04-09
Add snapcraft.yaml and initialization command

markshuttle:
Hello

As per the topic in Discussions, this PR adds initial snapcraft packaging for Tuwunel. The goal is to make it much easier for users of Ubuntu and other distros like Debian that have optional snapd to get Tuwunel installed and then get automatic updates.

Please see the Discussion topic for more details.

Mark


jevolk:
Thank you for this. I'll work on some of these lint flakes.

PR #414, opened 2026-04-13
fix: normalize storage provider variant name S3 to s3 for consistency

yonzilch:
The storage provider configuration currently uses inconsistent casing for variant names:

- `[global.storage_provider.<ID>.local]` — lowercase
- `[global.storage_provider.<ID>.S3]` — uppercase

This inconsistency is confusing and also causes practical issues when configuring via environment variables, since Tuwunel lowercases env var keys during parsing, making the `S3` variant unreachable through environment variables (users get `expected one of local, S3, None`).

In log it shows like:
Error: There was a problem with your configuration file: unknown variant: found `s3`, expected one of `local`, `S3`, `None` for key "STORAGE_PROVIDER.MEDIA_S3.S3" in `TUWUNEL_` environment variable(s)

**Proposed change:** Rename the `S3` variant to `s3` to match the convention established by `local`.

This would also fix environment variable configuration via `TUWUNEL_STORAGE_PROVIDER__<ID>__S3__*` which currently fails due to the case mismatch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5583,6 +5583,7 @@ dependencies = [
  "maplit",
  "nix 0.31.2",
  "num-traits",
+ "num_cpus",
  "object_store",
  "rand 0.8.5",
  "regex",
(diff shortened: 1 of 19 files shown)


jevolk:
I apologize for the CI and I will try to re-run this past any flakes. This should be a bit easier once we get this merged and you're made official :)

PR #426, opened 2026-04-21
feat: add configurable ip_source for reverse proxy and cloudflared deployments

theredspoon:
## Problem

When tuwunel runs behind a reverse proxy (nginx, Caddy) or Cloudflare Tunnel (`cloudflared`), the TCP peer address seen by tuwunel is the proxy's loopback address (`127.0.0.1`). The `axum_client_ip` layer is hardcoded to `SecureClientIpSource::ConnectInfo`, so all clients appear as `127.0.0.1` for any logging or security tooling that reads tuwunel logs (e.g. CrowdSec, fail2ban).

The nginx reverse proxy docs already include `proxy_set_header X-Forwarded-For $remote_addr`, but tuwunel has no way to tell its layer to read that header.

## What this PR does

Adds an `ip_source` config option (default: `"connect_info"`, preserving existing behaviour) that selects the `SecureClientIpSource` variant installed by the router layer at startup.

```toml
# Direct access (default — no change to existing deployments)
# ip_source = "connect_info"

# Behind nginx or Caddy
ip_source = "rightmost_x_forwarded_for"

# Behind Cloudflare Tunnel (cloudflared)
ip_source = "cf_connecting_ip"
```

A startup warning is logged when a header-based source is configured. Changes to `ip_source` during hot reload are rejected (restart required — the layer stack is not rebuilt on reload).

## Known limitation

Handlers that use `InsecureClientIp` (most API endpoints) scan headers in their own fixed order regardless of `SecureClientIpSource`. This PR wires the configured source into the layer correctly; migrating handlers from `InsecureClientIp` to `SecureClientIp` is a separate, broader change that touches ~22 files and warrants its own PR.

## Changes

- `src/core/config/mod.rs` — add `pub enum IpSource` (8 variants, `#[serde(rename_all = "snake_case")]`, `PartialEq`, `ConnectInfo` default) and `ip_source: IpSource` field
- `src/core/config/check.rs` — warn at startup when header-based source is configured; reject `ip_source` changes during hot reload
- `src/router/layers.rs` — replace hardcoded `SecureClientIpSource::ConnectInfo.into_extension()` with `ip_source_layer(&config.ip_source)` backed by exhaustive match
- `tuwunel-example.toml` — add commented `ip_source` entry with all supported values documented

## Supported values

| Value | Header / source | Use when |
| --- | --- | --- |
| `connect_info` | TCP peer address | Direct access (default) |
| `rightmost_x_forwarded_for` | `X-Forwarded-For` (rightmost) | nginx / Caddy |
| `rightmost_forwarded` | RFC 7239 `Forwarded` (rightmost) | RFC-compliant proxies |
| `x_real_ip` | `X-Real-IP` | nginx with `proxy_set_header X-Real-IP` |
| `cf_connecting_ip` | `CF-Connecting-IP` | Cloudflare or cloudflared |
| `true_client_ip` | `True-Client-IP` | Cloudflare Enterprise / Akamai |
| `fly_client_ip` | `Fly-Client-IP` | Fly.io |
| `cloudfront_viewer_address` | `CloudFront-Viewer-Address` | AWS CloudFront |

## No breaking changes

Default is `"connect_info"`. Existing deployments behave identically.

### DIFF
--- a/src/core/config/check.rs
+++ b/src/core/config/check.rs
@@ -3,7 +3,7 @@ use std::env::consts::OS;
 use either::Either;
 use itertools::Itertools;
 
-use super::{DEPRECATED_KEYS, IdentityProvider};
+use super::{DEPRECATED_KEYS, IdentityProvider, IpSource};
 use crate::{Config, Err, Result, debug, debug_info, error, warn};
 
 /// Performs check() with additional checks specific to reloading old config
@@ -16,6 +16,13 @@ pub fn reload(old: &Config, new: &Config) -> Result {
 			"server_name",
 			"You can't change the server's name from {:?}.",
 			old.server_name
+		));
+	}
+
+	if new.ip_source != old.ip_source {
+		return Err!(Config(
+			"ip_source",
+			"ip_source cannot be changed at runtime; restart the server to apply this change."
 		));
 	}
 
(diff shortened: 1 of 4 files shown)


theredspoon:
Withdrawing to reconsider scope. The config plumbing added here is a no-op until handlers are migrated from InsecureClientIp to SecureClientIp — those two changes belong together. Will resubmit as two PRs: first migrating handlers, then landing ip_source configuration on top.

PR #428, opened 2026-04-21
Add ClientIp extractor and migrate handlers

theredspoon:
## Summary

Introduces a tuwunel-internal `ClientIp` extractor in `src/api/client_ip.rs` and migrates all 22 API handler call sites away from `axum_client_ip::InsecureClientIp`.

This change is plumbing only. It is designed to be a no-op for every existing deployment:

- The extractor default branch delegates to `InsecureClientIp`, preserving today's header scan chain and `ConnectInfo` fallback exactly.
- A new marker, `ConfiguredIpSource`, gates the `SecureClientIp` branch. Nothing in this change installs that marker; #429 installs it only when `ip_source` is configured.
- The existing `SecureClientIpSource::ConnectInfo` extension installed by `src/router/layers.rs` is intentionally untouched and ignored by this extractor, so Unix-socket deployments (#310) remain unaffected.

## Two-PR series

This is the first PR in a two-PR effort for configurable, spoofing-resistant client IP resolution:

- **#428 (this PR):** introduces the internal `ClientIp` extractor and migrates handlers to use it, while preserving current behavior by default.
- **#429:** builds on this extractor by adding the optional `ip_source` config field, startup warnings for trusted-proxy header sources, reload protection, router-layer installation of `ConfiguredIpSource`, and deployment documentation.

Keeping this split lets reviewers verify the mechanical extractor migration independently from the user-facing configuration and documentation work. #429 should land after this PR.

## Test plan

- [x] `cargo check -p tuwunel_api`
- [x] `cargo clippy -p tuwunel_api --no-deps -- -D warnings`
- [x] `rustup run nightly rustfmt --check` on touched files
- [x] `cargo test -p tuwunel_api client_ip`
- [x] Unit tests in `src/api/client_ip.rs` cover fallback and configured paths, malformed headers, IPv6, XFF priority, configured-header absence, and the auto-installed `SecureClientIpSource` regression.
- [ ] Manual smoke: start tuwunel, hit `/_matrix/client/versions`, confirm tracing emits the connecting IP as today.

PR #429, opened 2026-04-21
Add configurable ip_source for client IP resolution.

### LINKED ISSUE #427: Client IP spoofing: InsecureClientIp in all handlers allows forged IPs; add configurable ip_source
## Problem

All API handlers use \`InsecureClientIp\` from \`axum-client-ip\`. This extractor scans headers in a fixed order and takes the **leftmost** \`X-Forwarded-For\` value. The leftmost XFF value is client-controlled — any client can send \`X-Forwarded-For: 1.2.3.4\` and tuwunel will log, rate-limit, and feed security tooling against that forged address.

The library's own documentation describes \`InsecureClientIp\` as appropriate when you "prefer to sacrifice security for probably statistically better IP determination," citing geolocation as the intended use case. Rate limiting and security tooling are the explicit counter-examples.

Two concrete problems with the current scan order:

1. **Leftmost = client-controlled.** A client sets \`X-Forwarded-For: <any IP>\` and wins.
2. **\`CF-Connecting-IP\` is checked last**, after \`X-Forwarded-For\`. Cloudflare Tunnel deployments remain spoofable even though \`CF-Connecting-IP\` is trustworthy — a client can inject a forged XFF value that gets picked up first.

## Affected deployments

Any tuwunel instance behind a reverse proxy — nginx, Caddy, Cloudflare Tunnel, Fly.io, AWS CloudFront. This is the recommended production topology per the docs.

## Proposed solution

Add an optional \`ip_source\` config field. When absent, handlers continue using \`InsecureClientIp\` exactly as today — no operator action required, no behavior change. When set, a request middleware resolves the IP using \`SecureClientIp\` with the configured source and stores it in extensions; handlers read from a new \`ClientIp\` extractor backed by those extensions.

\`SecureClientIp\` reads from exactly one configured source and takes the **rightmost** value on multi-value headers. The rightmost value is appended by the proxy and cannot be overridden by the client. It is spoofing-resistant.

```toml
# Not set (default) — InsecureClientIp behavior unchanged, no migration required

# Direct connection; explicitly disable header scanning
ip_source = "connect_info"

# Behind nginx or Caddy
ip_source = "rightmost_x_forwarded_for"

# Behind Cloudflare Tunnel (cloudflared)
ip_source = "cf_connecting_ip"
```

Supported values map 1:1 to \`SecureClientIpSource\` variants already in the crate:

| Value | Header | Proxy |
|---|---|---|
| `connect_info` | TCP peer address only | Direct, or to explicitly disable header scanning |
| `rightmost_x_forwarded_for` | `X-Forwarded-For` rightmost | nginx, Caddy |
| `rightmost_forwarded` | `Forwarded` rightmost | RFC 7239 proxies |
| `x_real_ip` | `X-Real-IP` | nginx |
| `cf_connecting_ip` | `CF-Connecting-IP` | Cloudflare, cloudflared |
| `true_client_ip` | `True-Client-IP` | Cloudflare Enterprise, Akamai |
| `fly_client_ip` | `Fly-Client-IP` | Fly.io |
| `cloudfront_viewer_address` | `CloudFront-Viewer-Address` | AWS CloudFront |

## ClientIp extractor design

The \`ClientIp\` wrapper extractor branches at request time:

- If \`SecureClientIpSource\` is present in extensions (i.e. \`ip_source\` is configured), delegate to \`SecureClientIp\` with that source.
- If not present, fall back to \`InsecureClientIp\` — preserving today's behavior exactly, including its header fallback chain.

This means Unix socket deployments are handled correctly: when \`ip_source\` is absent, the fallback to \`InsecureClientIp\`'s header scan remains in place, which matters because Unix sockets cannot provide \`ConnectInfo\` (see #310).

Note: \`SecureClientIpSource::ConnectInfo\` is already installed as an extension in \`src/router/layers.rs\` today (hardcoded). The infrastructure for \`SecureClientIp\` to resolve in handlers already exists — this proposal makes the source configurable rather than adding a new dependency.

## Trade-offs

**Security improvement is opt-in.** Operators who do not set \`ip_source\` keep existing behavior. This is intentional — the alternative is breaking all proxy deployments on update.

**\`ip_source\` is restart-required.** The router layer stack is built once at startup and is not rebuilt on config reload. Hot-reloading \`ip_source\` will be rejected with an error.

**Hard failure when configured header is absent.** \`SecureClientIp\` returns an error rather than falling back to the next available source. An operator who sets \`cf_connecting_ip\` with a misconfigured proxy will see errors rather than a degraded fallback. A startup warning when \`ip_source\` is set to a header-based value can prompt operators to verify their proxy configuration.

## Benefits once adopted

- CrowdSec HTTP scenarios and fail2ban see real client IPs behind all supported proxy topologies
- Rate limiting cannot be bypassed by header injection
- First-class support for Cloudflare Tunnel, Fly.io, and CloudFront — deployments where the correct header currently loses to a spoofable XFF in the scan order

## Planned contributions

I intend to submit two PRs against this issue:

1. **Handler migration** — introduce the \`ClientIp\` extractor with the two-mode fallback logic described above, and migrate all handlers from \`InsecureClientIp\` to \`ClientIp\`. Default behavior is unchanged for all existing deployments.

2. **Configurable \`ip_source\`** — add the \`ip_source\` config field (typed enum, serde \`rename_all = "snake_case"\`), wire it into the layer stack, add a startup warning for header-based sources, and reject \`ip_source\` changes during hot reload. This PR builds on PR 1 and is where the spoofing protection becomes available to operators.

theredspoon:
## Summary

Depends on #428 and should land after it.

Adds an optional `ip_source` config field for selecting how tuwunel resolves the client IP used by rate limiting, logging, and security tooling.

When `ip_source` is unset, no `ConfiguredIpSource` extension is installed, so the `ClientIp` extractor from #428 keeps the existing `InsecureClientIp` fallback behavior. When `ip_source` is set, the router installs a `ConfiguredIpSource` marker that maps to the selected `SecureClientIpSource` variant.

This also:

- rejects runtime reload changes to `ip_source`, because the router layer stack is built at startup;
- warns at startup when a header-based source is configured;
- documents `ip_source` in `tuwunel-example.toml` with a valid commented example;
- surfaces operator guidance in the generic deployment page and Caddy, Nginx, and Traefik reverse-proxy pages;
- calls out that Unix-socket deployments should leave `ip_source` unset rather than setting `connect_info`;
- teaches the config example generator a hidden `config-example:` directive, so fields can document an unset runtime default while emitting a valid example value;
- removes `syn` clone assumptions in the macro helper path so the touched macro crate remains directly testable.

## Macro note

`tuwunel-example.toml` is generated by `config_example_generator`, so a hand-edited example value would be overwritten on build. The new hidden `config-example:` directive lets `ip_source` keep its honest runtime default (`unset` / `None`) while generating the valid sample `#ip_source = "connect_info"`.

The small `src/macros/mod.rs` and `src/macros/implement.rs` changes remove existing `syn` clone assumptions that surfaced when adding direct tests for the generator behavior. They are included so `tuwunel_macros` can be checked, tested, and linted directly with this PR.

## Before / after

Before:

- client IP resolution followed the existing fallback behavior from `InsecureClientIp`;
- deployments behind trusted proxies had no typed config option for selecting a spoofing-resistant source;
- changing the intended source at runtime had no explicit guard because no source setting existed.

After:

- unset `ip_source` preserves the previous behavior;
- `ip_source = "connect_info"` uses the TCP peer address;
- header-based values such as `rightmost_x_forwarded_for` and `cf_connecting_ip` opt into the matching secure source and log a startup warning reminding operators to use a trusted reverse proxy;
- deployment docs explain which source to use for direct TCP, Unix sockets, and trusted reverse proxies;
- reload rejects `ip_source` changes until restart.

Closes #427.

## Test plan

- `git diff --check`
- `rustup run nightly rustfmt --check src/macros/config.rs src/macros/implement.rs src/macros/mod.rs src/core/config/mod.rs src/core/config/check.rs src/router/layers.rs`
- `cargo test -p tuwunel_macros`
- `cargo test -p tuwunel_core --no-default-features -- config::tests`
- `cargo test -p tuwunel_router --no-default-features -- layers::tests`
- `cargo check -p tuwunel_macros`
- `cargo clippy -p tuwunel_macros -- -D warnings`
- `cargo check -p tuwunel_core --no-default-features`
- `cargo check -p tuwunel_router --no-default-features`
- `cargo clippy -p tuwunel_core --no-default-features -- -D warnings`
- `cargo clippy -p tuwunel_core --no-default-features --tests -- -D warnings`
- `cargo clippy -p tuwunel_router --no-default-features -- -D warnings`
- `cargo clippy -p tuwunel_router --no-default-features --tests -- -D warnings`
- `lychee docs/deploying/generic.md docs/deploying/reverse-proxy-caddy.md docs/deploying/reverse-proxy-nginx.md docs/deploying/reverse-proxy-traefik.md` (31 links checked, 0 errors, 3 redirects)
- Confirmed the example remains `#ip_source = "connect_info"` and does not regenerate to `#ip_source = unset`.

On this local macOS checkout, router-targeted commands used `CPLUS_INCLUDE_PATH=/Library/Developer/CommandLineTools/SDKs/MacOSX15.2.sdk/usr/include/c++/v1` because the local `clang++` installation did not find libc++ headers by default.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4392,9 +4392,9 @@ checksum = "f87165f0995f63a9fbeea62b64d10b4d9d8e78ec6d7d51fb2125fda7bb36788f"
 
 [[package]]
 name = "rustls-webpki"
-version = "0.103.12"
+version = "0.103.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8279bb85272c9f10811ae6a6c547ff594d6a7f3c6c6b02ee9726d1d0dcfcdd06"
+checksum = "61c429a8649f110dddef65e2a5ad240f747e85f7758a6bccc7e5777bd33f756e"
 dependencies = [
  "aws-lc-rs",
  "ring",
--- a/docs/deploying/generic.md
+++ b/docs/deploying/generic.md
@@ -166,6 +166,26 @@ Regardless of which reverse proxy you choose, you will need to:
    - Port 443 (HTTPS) for client-server API
    - Port 8448 for federation (if federating with other homeservers)
 
+### Client IP source
+
+Set `ip_source` when you want Tuwunel to use a spoofing-resistant client IP
+source for rate limiting, logging, and security tooling. Leave it unset to keep
+the legacy fallback behavior.
+
+Use `ip_source = "connect_info"` only when Tuwunel accepts direct TCP
+connections and should use the TCP peer address. Do not use `connect_info` for
+Unix-socket deployments; leave `ip_source` unset there.
+
+If Tuwunel is behind a trusted reverse proxy, set `ip_source` to match the
+header that proxy controls. Caddy, Nginx, and Traefik usually use
+`ip_source = "rightmost_x_forwarded_for"`. Cloudflare and cloudflared
+deployments can use `ip_source = "cf_connecting_ip"` when Cloudflare supplies
+that header.
+
+Only use header-based values when clients cannot connect to Tuwunel directly.
+If clients can reach Tuwunel without going through the trusted proxy, they can
+send forged forwarding headers and choose the IP address Tuwunel sees.
+
 See the following spec pages for more details on well-known files:
 - [`/.well-known/matrix/server`](https://spec.matrix.org/latest/client-server-api/#getwell-knownmatrixserver)
 - [`/.well-known/matrix/client`](https://spec.matrix.org/latest/client-server-api/#getwell-knownmatrixclient)
--- a/docs/deploying/reverse-proxy-caddy.md
+++ b/docs/deploying/reverse-proxy-caddy.md
@@ -28,6 +28,12 @@ your.server.name, your.server.name:8448 {
 - Sets all necessary reverse proxy headers correctly
 - Routes all traffic to Tuwunel listening on `localhost:8008`
 
+### Client IP source
+
+If Caddy is the only way clients can reach Tuwunel, set
+`ip_source = "rightmost_x_forwarded_for"` in `tuwunel.toml`. If you use the
+Unix-socket `reverse_proxy` target, leave `ip_source` unset instead.
+
 That's it! Just start and enable the service and you're set.
 
 ```bash
--- a/docs/deploying/reverse-proxy-nginx.md
+++ b/docs/deploying/reverse-proxy-nginx.md
@@ -84,6 +84,9 @@ server {
 
 - **Replace `matrix.example.com`** with your actual server name
 - **`client_max_body_size`**: Must match or exceed `max_request_size` in your `tuwunel.toml`
+- **`ip_source`**: If Nginx is the only way clients can reach Tuwunel, set
+  `ip_source = "rightmost_x_forwarded_for"` so Tuwunel uses the trusted
+  `X-Forwarded-For` value
 - **Do NOT use `$request_uri`** in `proxy_pass` - while some guides suggest this, it's not necessary for Tuwunel and can cause issues
 - **IPv6**: The `listen [::]:443` and `listen [::]:8448` lines enable IPv6 support. Remove them if you don't need IPv6
 
--- a/docs/deploying/reverse-proxy-traefik.md
+++ b/docs/deploying/reverse-proxy-traefik.md
@@ -76,7 +76,15 @@ http:
                     - url: "http://tuwunel:6167"
                 passHostHeader: true
 ```
+
+### Client IP source
+
+If Traefik is the only way clients can reach Tuwunel, set
+`ip_source = "rightmost_x_forwarded_for"` in `tuwunel.toml` so Tuwunel uses the
+trusted `X-Forwarded-For` value.
+
 ### Federation
+
 If you will use a .well-known file you can use traefik to redirect .well-known/matrix to tuwunel built-in .well-known file.
 
 replace the rule in either of the methods from
--- a/src/api/client/account.rs
+++ b/src/api/client/account.rs
@@ -1,5 +1,4 @@
 use axum::extract::State;
-use axum_client_ip::InsecureClientIp;
 use futures::{FutureExt, StreamExt};
 use ruma::api::client::account::{
 	ThirdPartyIdRemovalStatus, change_password, deactivate, get_3pids,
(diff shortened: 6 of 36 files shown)


theredspoon:
Update after the restack:

- #429 is now restacked on top of #428's amended audit-fix commit.
- Current #428 base commit: `b0b15dca4e3e5fb2b0e323bab15e98e643ac2207`.
- Current #429 head: `ed9fd3d55057f2c8d08b6b241b6b402a02b3fc60`.
- Local stack shape was verified as one #428 commit over `main`, and one #429 commit over #428.

Local validation performed:

- Ran the Docker `clippy` bake target against this PR stack with the same relevant CI matrix shape: `cargo_profiles=["test"]`, `feat_sets=["all"]`, `rust_toolchains=["nightly"]`, Debian `testing-slim`, Rust target `x86_64-unknown-linux-gnu`, sys target `x86_64-v1-linux-gnu`. It completed successfully with the CI `-D warnings` Clippy path. Local note: I set `cargo_installs=cargo-chef` for that Clippy run so the build skipped unrelated cargo helper installs and reached the lint step.
- Ran `cargo audit --stale --deny yanked --deny unsound --deny unmaintained --deny warnings --color never` locally with `cargo-audit 0.22.1`; it exited successfully on this restacked head, inheriting #428's `rustls-webpki 0.103.13` lockfile fix.

Constraints / caveats:

- The full Docker `audit` bake target did not complete locally because the Docker/Colima environment exhausted host disk space and then rustup hit `Input/output error (os error 5)` before the advisory scan. I used the host `cargo-audit` command above to verify the actual advisory condition instead.
- Fresh GitHub Actions runs after the force-push are currently `action_required` with `jobs: []`, so upstream CI has not actually rerun yet and likely needs maintainer approval/re-run for the forked PR.

theredspoon:
#429 includes all of #428 plus the `ip_source` follow-up.

It passed the same `rust-sdk-integ` jobs that failed on #428, so those failures don’t look related to #428’s code.

Given that, my preferred path would be to merge #429 and close #428 as superseded.

If you’d rather preserve the two-PR split, I’d appreciate a rerun of #428 so you can merge them in order if it comes back green.

PR #449, opened 2026-05-08
fix: always include `bump_stamp`

lhjt:
## Summary

Fix sliding sync room updates so `bump_stamp` is not accidentally omitted when the response only includes part of the room timeline.

Before this change, tuwunel calculated `bump_stamp` by looking only at the timeline events it was about to send back to the client. That meant `bump_stamp` could be missing even when the room had new activity, if the relevant event was not included in that particular response.

I ran into this while using the Matrix Rust SDK: the SDK ingests MSC4186 `bump_stamp` as the room’s recency stamp, which clients commonly use to order the room list by recent activity. When tuwunel omits it, the SDK can receive the new timeline event while still leaving the room’s recency value stale, so the room may not move to its expected position in the list.

Should hopefully fix the issue in Element X and other clients where a room with a recent update just wouldn't appear in the sidebar until a new message arrives (in some cases).

## What changed

Instead of calculating `bump_stamp` from the returned timeline slice, tuwunel now looks at the room activity since the last sync and finds the newest event that should bump the room.

The timeline response is still limited the same way as before. The only change is that the room’s “recent activity” marker no longer depends on which timeline events happened to be included in the response.

## Why this matters

A client could previously receive the new message but not receive the matching `bump_stamp`.

That could make the room look stale in the sidebar even though it had new activity.

### DIFF
--- a/src/api/client/sync/v5/rooms.rs
+++ b/src/api/client/sync/v5/rooms.rs
@@ -1,53 +1,32 @@
-use std::{
-	cmp::Ordering,
-	collections::{BTreeMap, HashSet},
-};
+mod bump_stamp;
+mod heroes;
+
+use std::collections::{BTreeMap, HashSet};
 
 use futures::{
 	FutureExt, StreamExt, TryFutureExt, TryStreamExt,
 	future::{join, join3, join4},
 };
 use ruma::{
-	JsOption, MxcUri, OwnedMxcUri, OwnedRoomId, RoomId, UInt, UserId,
-	api::client::sync::sync_events::{
-		UnreadNotificationsCount,
-		v5::{DisplayName, response, response::Heroes},
-	},
-	events::{
-		StateEventType,
-		TimelineEventType::{
-			self, Beacon, CallInvite, PollStart, RoomEncrypted, RoomMessage, Sticker,
-		},
-		room::member::MembershipState,
-	},
+	JsOption, OwnedRoomId, UserId,
+	api::client::sync::sync_events::{UnreadNotificationsCount, v5::response},
+	events::{StateEventType, TimelineEventType, room::member::MembershipState},
 };
 use tuwunel_core::{
 	Result, at, err, error, is_equal_to,
 	itertools::Itertools,
-	matrix::{
-		Event, StateKey,
-		pdu::{PduCount, PduEvent},
-	},
+	matrix::{Event, StateKey, pdu::PduCount},
 	ref_at,
 	utils::{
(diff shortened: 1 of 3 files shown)


jevolk:
### Servers

| Server | Window | Bumpable types | Includes `m.room.create` | Self-membership | Non-join carve-out | Filters redacted | Value |
|---|---|---|---|---|---|---|---|
| Tuwunel (before) | timeline slice only | 6 | no | yes (when in slice) | no | yes | `PduCount` |
| Tuwunel (after) | full `(roomsince, next_batch]` reverse-walk | 6 | no | yes | no | yes | `PduCount` |
| Synapse | full up to `to_token` (DB query, all events) | 7 | yes | no | yes (uses membership event stream pos) | no | stream_ordering, negatives clamped to 0 |
| Continuwuity | timeline slice only | 6 | no | no | no | no | `origin_server_ts` |
| Conduit | timeline slice only | any event (no filter) | yes (implicit) | yes (any user, no filter) | no | no | `origin_server_ts` |
| Zendrite | timeline first, then DB fallback over full room history | 7 | yes | no | no | no | heterogeneous: `OriginServerTS` or stream position |

### Clients

| Client | Parses `bump_stamp`? | Stored as | Omission handling | Used for room-list sort | Role in sort | Citation |
|---|---|---|---|---|---|---|
| matrix-rust-sdk | yes | `RoomInfo.recency_stamp: Option<u64>` | preserves prior value (covered by test) | yes | fallback when neither room has `LatestEventValue` ts | `base/src/response_processors/room/msc4186/mod.rs:363`; sort at `ui/src/room_list_service/sorters/recency.rs:36-100`; preservation test `base/src/sliding_sync.rs:1390-1404` |
| matrix-js-sdk | yes | `Room.bumpStamp: number \| undefined`, exposed via `getBumpStamp()` | no preservation logic in SDK; setter overwrites only when present | exposes only; consumer decides | n/a (consumer-side) | `sliding-sync.ts:112`, `sliding-sync-sdk.ts:612-613`, `models/room.ts:341,1654-1664` |
| Element-Web | via `room.getBumpStamp()` | (delegated to js-sdk) | falls back to last-timeline-event timestamp when getter returns falsy | yes | primary when present | `room-list-v3/skip-list/sorters/utils/getLastTimestamp.ts:41-43`, `room-list/algorithms/tag-sorting/RecentAlgorithm.ts:74-76`; test pin `getLastTimestamp-test.ts:66-89` |
| Out-Of-Your-Element | typed but unused (AS bridge calls SSS only for invite state) | not stored | n/a | no (AS bridge, no room list UI) | n/a | `src/types.d.ts:463`, `src/matrix/api.js:202-224` |

jevolk (approved):

PR #461, opened 2026-05-15
README.md: add Gentoo ebuild to "Getting Started" section.

NiklausHofer:
Guru is Gentoo's community repo, sort of what AUR is to Arch I guess.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ Switzerland 🇨🇭 where it is currently deployed for citizens.
 - Arch package available as [tuwunel](https://aur.archlinux.org/packages/tuwunel).
 - Nix package available as [`matrix-tuwunel`](https://search.nixos.org/packages?query=matrix-tuwunel) and NixOS module available as [`services.matrix-tuwunel`](https://search.nixos.org/options?query=services.matrix-tuwunel).
 - Alpine package available as [tuwunel](https://pkgs.alpinelinux.org/package/edge/testing/x86_64/tuwunel).
+- Gentoo ebuild available in Guru as [net-im/tuwunel](https://gitweb.gentoo.org/repo/proj/guru.git/tree/net-im/tuwunel).
 - Ansible playbook available as [matrix-docker-ansible-deploy](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-tuwunel.md).
 
 **1.** [Configure](https://matrix-construct.github.io/tuwunel/configuration.html) by


x86pup:
Hi, I was told that there may be a concern with this ebuild about this line:

```
SRC_URI="
	https://github.com/matrix-construct/${PN}/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz
	https://vimja.cloud/public.php/dav/files/z59eKDyLFokW2KK/${CATEGORY}/${PN}/${P}-vendor.tar.xz
"
```
 

> Not wrong per se but really weird, I'm not sure how it got accepted into guru
usually rust crates are downloaded directly from crates.io, this one is using some random webdav to get a tarball of already vendored crates, so whoever is in control of that tarball can add malicious code to it

Is there a reason the ebuild needs to pull from your website? I think I would be concerned about it using infrastructure that only you control for this reason too. There's been many supply chain compromises and such lately so this is fairly concerning. None of us know anything about ebuilds so I didn't see the problem until I asked some Gentoo friends about it.

Here's an example of how to do it properly: https://codeberg.org/gentoo/guru/src/branch/master/x11-wm/leftwm/leftwm-0.5.4.ebuild

NiklausHofer:
> Here's an example of how to do it properly: https://codeberg.org/gentoo/guru/src/branch/master/x11-wm/leftwm/leftwm-0.5.4.ebuild

That's what I tried at first. Unfortunately, it was not able to handle the async-channel dependency correctly. I've asked about it in the Gentoo forums as well, see here: https://forums.gentoo.org/viewtopic.php?p=8883167&hilit=tuwunel#p8883167

The Vendor Tarball option is [described in the Gentoo wiki](https://wiki.gentoo.org/wiki/Writing_Rust_ebuilds#Using_a_vendor_tarball_like_in_Go_ebuilds), too. It's not something I came up with on my own. In fact, the `pkgcheck` utility used for ebuild quality control will complain when the list of crates gets really long and suggest switching to a vendor tarball. I think it did that for the tuwunel ebuild too.

> Is there a reason the ebuild needs to pull from your website? I think I would be concerned about it using infrastructure that only you control for this reason too. There's been many supply chain compromises and such lately so this is fairly concerning.

I absolutely understand the concerns and I don't think it's a great option, either. Not just because of security concerns, but for availability reasons, too. The Nextcloud instance in my living room doesn't exactly provide great uptime :/

On the other hand, I should maybe point out that using [vendor tarballs is now the preferred way of packaging Go software](https://wiki.gentoo.org/wiki/Writing_go_Ebuilds#Packaging_the_dependencies). Because of that, a whole number of packages in Guru pull tarballs from random websites. 

> None of us know anything about ebuilds so I didn't see the problem until I asked some Gentoo friends about it.

Could you maybe point some of them at the forum entry linked above? I would be very grateful for any hints on how to better handle this!

x86pup:
I pointed them to the forum post and maybe they'll have insight. Though I suppose if this is a documented and supposedly recommended way to do it then I guess it will suffice, but would be better to figure out long-term if this tarball can be hosted elsewhere..

NiklausHofer:
Back to let you know that I'm taking your concerns and your recommendations seriously. After a lot of trial and error, I've finally managed to drop the vendor tarball. The ebuild no longer downloads artifacts from my own infra:

https://github.com/gentoo/guru/blob/master/net-im/tuwunel/tuwunel-1.7.1.ebuild

Granted, there is still a tarball involved, but that is created transparently by a set-up I have no control over: https://github.com/gentoo-crate-dist/tuwunel

PR #477, opened 2026-06-03
fix: use skip_while + take_while for room-scoped timestamp scan

lingbohome:
### Problem

`/timestamp_to_event` (MSC3030) returns `M_NOT_FOUND` for valid searches. 

### Root Cause

`pdu_ids_near_ts` scans the `roomid_ts_pducount` index keyed by `(RoomId, timestamp)`. Keys from different rooms are interleaved. When the scan starts at a position belonging to another room, `ready_try_take_while` **terminates** the stream on the first non-matching key, returning no results.

For prefix-based scans (`starts_with`) `take_while` is correct because non-matching keys are contiguous. For interleaved composite keys, the scan must skip over non-matching entries first.

### Fix

Replace the single `ready_try_take_while` with `ready_try_skip_while` (skip to target room) + `ready_try_take_while` (stop when room changes). This correctly bounds the stream to only keys belonging to the target room.

### DIFF
--- a/src/service/rooms/timeline/pdus.rs
+++ b/src/service/rooms/timeline/pdus.rs
@@ -90,7 +90,8 @@ pub fn pdu_ids_near_ts(
 				| Forward => Left(self.db.roomid_ts_pducount.stream_from(&key)),
 				| Backward => Right(self.db.roomid_ts_pducount.rev_stream_from(&key)),
 			}
-			.ready_try_filter(move |((room_id_, _), _): &KeyVal<'_>| room_id == *room_id_)
+			.ready_try_skip_while(move |((room_id_, _), _): &KeyVal<'_>| Ok(room_id != *room_id_))
+				.ready_try_take_while(move |((room_id_, _), _): &KeyVal<'_>| Ok(room_id == *room_id_))
 			.map_ok(move |((_, ts), count)| {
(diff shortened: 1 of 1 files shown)


jevolk:
Unfortunately the timestamp-to-event/jump-to-date implementation has a large list of problems which weren't initially uncovered. Fixing them appears to require a database migration at this point, changing the column's schema to fix some fundamental shortcomings. The issue being solved here is probably among those as well, so it looks like I can merge this and then apply the other fixes afterward.

Given there's no quick fix here, it might be several days before I get to this to address it all at once as a unit.

lingbohome:
Thanks for the review.

We've been running this patch against a fork build and it works correctly — newly persisted events are searchable via /timestamp_to_event. Only events that pre-date the Tuwunel upgrade (before the roomid_ts_pducount index existed) remain unfindable, which is expected and not something this PR aims to solve.

The reason we built our own fork image instead of relying on the project CI is that the docker/bake.hcl build system is too heavy for a personal fork without self-hosted runners (the Main workflow never starts). We ended up with a minimal Dockerfile just for x86_64. If there's a simpler way to build a single-arch image from a fork, that advice would be helpful.

That said, please merge this when you're ready — we'll rebase and rebuild once your larger fix lands.

PR #487, opened 2026-06-15
[nginx] Use one server block for both 443 and 8448

Daniel15:
There's no need for two identical server blocks that differ only by port. We can instead use one server block to serve both ports.


jevolk (approved):

jevolk:
Nightly lints unrelated; disposition is still merge.

PR #511, opened 2026-07-14
users: preserve zero one-time-key counts

basnijholt:
While testing encrypted MatrixRTC voice calls for MindRoom agents, I hit a failure once the agent's signed Curve25519 one-time-key pool had been fully consumed. Tuwunel builds the count map from rows currently in the store; after the last key is claimed the map becomes empty, and Ruma omits the empty `device_one_time_keys_count` from `/sync`. Clients that distinguish an omitted count (no update / unknown) from an explicit zero can then leave an already-initialized device without replenishing its pool, so fresh peers cannot establish Olm sessions.

This keeps `signed_curve25519` visible at zero after a pool is drained, matching Synapse and the algorithm clients use for Olm one-time keys. Existing non-zero counts and any other algorithms already present in the map are preserved. The missing-MSC4225-column path still returns an empty map, because that node cannot observe the authoritative pool and should not manufacture a zero count.

Although we found this through encrypted MatrixRTC media-key exchange, the issue is general to any E2EE device whose one-time-key pool reaches zero. For context, this surfaced during the [MindRoom MatrixRTC voice-call work](https://github.com/mindroom-ai/mindroom/pull/1452); the client-side half of the drained-pool recovery is described in [mindroom-ai/mindroom#1499](https://github.com/mindroom-ai/mindroom/pull/1499).

Verified on top of current `main`:

- `cargo +nightly fmt --all -- --check`
- `cargo test -p tuwunel_service one_time_key_counts_` — 2 tests pass
- `cargo clippy -p tuwunel_service --tests -- -D warnings`

### DIFF
--- a/src/service/users/keys.rs
+++ b/src/service/users/keys.rs
@@ -310,51 +310,17 @@ pub async fn count_one_time_keys(
 
 /// Keep zero-count algorithms visible to clients after an OTK pool is drained.
 ///
-/// An empty map is omitted from `/sync` by ruma. Some clients, including nio,
-/// interpret an omitted count as "unknown" and therefore do not replenish a
-/// previously-uploaded Olm account. `curve25519` is retained for compatibility
-/// with clients which still require the legacy count in `/keys/upload`.
+/// An empty map is omitted from `/sync` by ruma. Some clients interpret an
+/// omitted count as "unknown" and therefore do not replenish a
+/// previously-uploaded Olm account. Only `signed_curve25519` is seeded,
+/// matching Synapse; clients do not maintain unsigned curve25519 keys.
 fn complete_one_time_key_counts(
 	mut counts: BTreeMap<OneTimeKeyAlgorithm, UInt>,
 ) -> BTreeMap<OneTimeKeyAlgorithm, UInt> {
 	counts
-		.entry(OneTimeKeyAlgorithm::from("curve25519"))
-		.or_default();
-	counts
 		.entry(OneTimeKeyAlgorithm::SignedCurve25519)
 		.or_default();
 	counts
-}
-
-#[cfg(test)]
-mod tests {
-	use super::*;
-
-	#[test]
-	fn empty_one_time_key_counts_include_supported_zeroes() {
-		let counts = complete_one_time_key_counts(BTreeMap::new());
-
-		assert_eq!(
-			counts.get(&OneTimeKeyAlgorithm::from("curve25519")),
-			Some(&UInt::from(0_u32))
-		);
-		assert_eq!(counts.get(&OneTimeKeyAlgorithm::SignedCurve25519), Some(&UInt::from(0_u32)));
-	}
-
-	#[test]
-	fn existing_one_time_key_counts_are_preserved() {
-		let mut counts = BTreeMap::new();
-		counts.insert(OneTimeKeyAlgorithm::from("curve25519"), UInt::from(11_u32));
-		counts.insert(OneTimeKeyAlgorithm::SignedCurve25519, UInt::from(17_u32));
-
-		let counts = complete_one_time_key_counts(counts);
-
-		assert_eq!(
-			counts.get(&OneTimeKeyAlgorithm::from("curve25519")),
-			Some(&UInt::from(11_u32))
-		);
-		assert_eq!(counts.get(&OneTimeKeyAlgorithm::SignedCurve25519), Some(&UInt::from(17_u32)));
-	}
 }
 
 /// MSC4225: drop the `excess` oldest rows for this `(user, device)`. Forward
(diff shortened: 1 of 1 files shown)


src/service/users/keys.rs
```diff
@@ -303,7 +305,56 @@ pub async fn count_one_time_keys(
...
+	mut counts: BTreeMap<OneTimeKeyAlgorithm, UInt>,
+) -> BTreeMap<OneTimeKeyAlgorithm, UInt> {
+	counts
+		.entry(OneTimeKeyAlgorithm::from("curve25519"))
+		.or_default();
+	counts
+		.entry(OneTimeKeyAlgorithm::SignedCurve25519)
+		.or_default();
+	counts
+}
```
jevolk, line 324: Seed `signed_curve25519` only. The legacy entry costs more than it buys: Complement's `TestUploadKeyIdempotencyOverlap` iterates `one_time_key_counts` expecting every entry to equal the uploaded count, so `curve25519: 0` beside `signed_curve25519: 3` flips a test our CI passes today. ruma has no unit variant for unsigned curve25519, so `OneTimeKeyAlgorithm::from("curve25519")` allocates a `PrivOwnedStr` on every call. And no client reads it: matrix-rust-sdk, matrix-js-sdk, nheko, matrix-dart-sdk, and mautrix-go consult only `signed_curve25519`, while upstream nio drops an explicit zero on a truthiness check and the mindroom-nio fix reads `signed_curve25519`, so even the nio lineage gets nothing from it. Synapse seeds only `signed_curve25519`.

```suggestion
/// Keep zero-count algorithms visible to clients after an OTK pool is drained.
///
/// An empty map is omitted from `/sync` by ruma. Some clients interpret an
/// omitted count as "unknown" and therefore do not replenish a
/// previously-uploaded Olm account. Only `signed_curve25519` is seeded,
/// matching Synapse; clients do not maintain unsigned curve25519 keys.
fn complete_one_time_key_counts(
	mut counts: BTreeMap<OneTimeKeyAlgorithm, UInt>,
) -> BTreeMap<OneTimeKeyAlgorithm, UInt> {
	counts
		.entry(OneTimeKeyAlgorithm::SignedCurve25519)
		.or_default();
	counts
}
```

src/service/users/keys.rs
```diff
@@ -303,7 +305,56 @@ pub async fn count_one_time_keys(
...
+
+		let counts = complete_one_time_key_counts(counts);
+
+		assert_eq!(
+			counts.get(&OneTimeKeyAlgorithm::from("curve25519")),
+			Some(&UInt::from(11_u32))
+		);
+		assert_eq!(counts.get(&OneTimeKeyAlgorithm::SignedCurve25519), Some(&UInt::from(17_u32)));
+	}
 }
```
jevolk: Move the test module to the end of the file; it lands between `count_one_time_keys` and `prune_one_time_keys` here, and the file runs another 430 lines. With the legacy entry dropped, the module becomes:

```rust
#[cfg(test)]
mod tests {
	use super::*;

	#[test]
	fn empty_one_time_key_counts_include_signed_zero() {
		let counts = complete_one_time_key_counts(BTreeMap::new());

		assert_eq!(counts.len(), 1);
		assert_eq!(counts.get(&OneTimeKeyAlgorithm::SignedCurve25519), Some(&UInt::from(0_u32)));
	}

	#[test]
	fn existing_one_time_key_counts_are_preserved() {
		let mut counts = BTreeMap::new();
		counts.insert(OneTimeKeyAlgorithm::from("curve25519"), UInt::from(11_u32));
		counts.insert(OneTimeKeyAlgorithm::SignedCurve25519, UInt::from(17_u32));

		let counts = complete_one_time_key_counts(counts);

		assert_eq!(
			counts.get(&OneTimeKeyAlgorithm::from("curve25519")),
			Some(&UInt::from(11_u32))
		);
		assert_eq!(counts.get(&OneTimeKeyAlgorithm::SignedCurve25519), Some(&UInt::from(17_u32)));
	}
}
```

jevolk (changes requested):
Thanks for running this down; the failure mode is real and reaches well past nio. We verified the client side: matrix-js-sdk and matrix-rust-sdk (Element Web and X, Fractal, NeoChat) and nheko all act only when a `signed_curve25519` entry is present in the map, so an omitted count after a drained pool never triggers replenishment, while an explicit zero does. Synapse has seeded `signed_curve25519: 0` into every count response since 2021 for the same reason, and that is the shape to match.

Two changes requested inline: seed only `signed_curve25519`, and move the unit test module to the end of the file.

src/service/users/keys.rs
basnijholt: Addressed in 007033cd: only signed_curve25519 is seeded now; existing entries remain unchanged.

src/service/users/keys.rs
basnijholt: Addressed in 007033cd: moved the test module to the end of the file and updated the empty-map assertion.

jevolk (approved):

basnijholt:
The failing nightly integration job looks unrelated to this change. `appservice_e2ee_transactions` fails during setup with `Duplicate id: voicebridge_off`, before the one-time-key count code runs; the stable integration job passed on the same SHA.

I do not have permission to rerun upstream Actions jobs. Could you rerun the failed job once the workflow finishes?

basnijholt:
All passes now, thanks! Also thanks for all the quick reviews and merges always @jevolk!

PR #518, opened 2026-07-18
fix: prevent stale read receipt updates

### LINKED ISSUE #516: Equal or older public read receipts are persisted and re-emitted instead of ignored
### Tuwunel version

1.8.2

### Installation method

Docker / OCI image

### Platform and OS

x86 OCI container

### Database origin

Created by Tuwunel

### What happened?

Tuwunel persists and re-emits a public `m.read` receipt even when the new receipt points to the same event, or an older event, than the receipt already stored for the same room, user, receipt type, and thread context.

Each submission receives a new receipt stream position and is sent again as an `m.receipt` EDU to interested appservices. A changed receipt timestamp appears to be sufficient to cause another update even though the user's read position has not advanced.

This can create an unbounded feedback loop with a bridge:

1. An acknowledgement on the remote network causes the bridge to post a Matrix read marker.
2. Tuwunel emits the resulting `m.receipt` to the bridge because `receive_ephemeral: true`.
3. The bridge handles that receipt and acknowledges the remote message.
4. The remote acknowledgement causes the bridge to post the same Matrix read marker again.
5. Tuwunel stores and emits it again.

In my scenario this sequence repeated 2,113 times before the bridge was stopped.

For comparison, Synapse ignores a receipt when its event ordering is equal to or older than the stored receipt. It only creates a new receipt stream entry when the read position advances:

https://github.com/element-hq/synapse/blob/5ed830b3b4c74c89d876cc07756c5d98a100cbed/synapse/storage/databases/main/receipts.py#L885-L946

I will be raising a PR to fix this issue 🙂

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### Relevant log output or configuration

```shell

```

lhjt:
### What does this PR do?

Fixes #516 

### Checklist

      comment.
      Tuwunel's Code of Conduct.

### DIFF
--- a/src/api/client/read_marker/read_markers.rs
+++ b/src/api/client/read_marker/read_markers.rs
@@ -29,19 +29,6 @@ pub(crate) async fn set_read_marker_route(
 ) -> Result<set_read_marker::v3::Response> {
 	let sender_user = body.sender_user();
 
-	if body.private_read_receipt.is_some() || body.read_receipt.is_some() {
-		// Route through the dispatcher so per-thread counts are also cleared;
-		// `/read_markers` predates MSC3771 and carries no thread field.
-		services
-			.pusher
-			.reset_notification_counts_for_thread(
-				sender_user,
-				&body.room_id,
-				&ReceiptThread::Unthreaded,
-			)
-			.await;
-	}
-
 	if let Some(event) = &body.fully_read {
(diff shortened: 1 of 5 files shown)


jevolk:
The only real problem with this one was the unbounded iteration and the empty receipt issue, otherwise the rest is optimal so this should be going in, sorry.

PR #520, opened 2026-07-19
fix(pusher): refresh badges after read updates

lhjt:
### What does this PR do?

Push badge counts were only sent to a push gateway as part of a normal event notification.

When a user read messages, tuwunel correctly reset the server-side unread counters, but it did not tell registered pushers that the count had changed. Clients (such as Element X) could therefore continue displaying a stale badge until another event happened to generate a push.

### Changes

Send a counts-only push update after an operation clears unread notification counts.

These updates are triggered when:

- A read receipt advances.
- A read marker advances.
- A locally sent event clears the sender’s unread count.

The counts-only payload contains the current global unread count and target device, but no event ID, room ID, content, sender, notification type, or alert tweaks. It can therefore update or clear the application badge without representing a new message notification.

### Checklist

      comment.
      Tuwunel's Code of Conduct.

### DIFF
--- a/src/api/client/read_marker/read_markers.rs
+++ b/src/api/client/read_marker/read_markers.rs
@@ -40,6 +40,9 @@ pub(crate) async fn set_read_marker_route(
 				&ReceiptThread::Unthreaded,
 			)
 			.await;
+		services
+			.pusher
+			.schedule_badge_update(sender_user.to_owned());
 	}
 
 	if let Some(event) = &body.fully_read {
--- a/src/api/client/read_marker/receipt.rs
+++ b/src/api/client/read_marker/receipt.rs
@@ -77,6 +77,9 @@ pub(crate) async fn create_receipt_route(
 			.pusher
 			.reset_notification_counts_for_thread(sender_user, &body.room_id, &body.thread)
 			.await;
+		services
+			.pusher
+			.schedule_badge_update(sender_user.to_owned());
 	}
 
 	match body.receipt_type {
--- a/src/database/maps.rs
+++ b/src/database/maps.rs
@@ -367,6 +367,11 @@ pub(super) static MAPS: &[Descriptor] = &[
 		..descriptor::RANDOM_SMALL
 	},
 	Descriptor {
+		name: "senderkey_lastbadge",
+		val_size_hint: Some(8),
+		..descriptor::RANDOM_SMALL
+	},
+	Descriptor {
 		name: "senderkey_pusher",
 		..descriptor::RANDOM_SMALL
 	},
--- a/src/main/tests/pusher_notify.rs
+++ b/src/main/tests/pusher_notify.rs
@@ -31,6 +31,7 @@ const APP_ID: &str = "app.tuwunel.test";
 const EVENT_ID: &str = "$push:remote.example";
 const SENDER: &str = "@alice:remote.example";
 const NOTIFY_PATH: &str = "/_matrix/push/v1/notify";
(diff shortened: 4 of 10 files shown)


lhjt:
Hi @jevolk I was wondering if you could shed some light on why this PR was closed, thanks!

jevolk:
I'm sorry but we tried and could not see a path forward for #518, #519 and #520. This one was the most workable of the three, but they all basically share the same trouble; extending also to #517 but it was small enough to force through.

**All of these PR's are huge. On average about 75% larger and more complicated than required to solve the problem at issue.**

- Start with #517 as an example: the problem is that presence updates need to be ignored from appservice users. That requires a single bit discriminator. One bit of information. That can take the form of a bool, a branch, etc. But #517 offered a [three-piece enum](https://github.com/matrix-construct/tuwunel/commit/b48ffbc7a4e6aaa6f9ef97b356518453f878d67e#diff-032d01274a79affbdbcf81cc9230da43e4a538544dc92b31f76ed996c035012bL32-L52) and arguments identifying entirely uninvolved variants determined for every callsite of `maybe_ping_presence`. We took the opportunity to make the callsites and the arguments more readable and forward-scalable, and I'm even guilty of bloating it with an instance of `Option<&RegistrationInfo>` rather than a plain `bool`. But we had to basically steamroll the PR.

- Now it gets bigger, take #519 where we're [removing](https://github.com/matrix-construct/tuwunel/commit/a32f80c09bd69366d4a32a2ead0533843037299c#diff-0246cdedc16ec974a6497cfaca535dd8542759b11b53049e5149595e5009b0a5L1-L119) entire [files](https://github.com/matrix-construct/tuwunel/commit/a32f80c09bd69366d4a32a2ead0533843037299c#diff-da949bec59604ec37109a15db9d8847327a285ef9064fa03acc88d2c8bcbe01cL1-L139) and still satisfying the same objective, fixing the bug. Unlike #517 we can't jam this one through because we can't tear apart sliding-sync at this large scale on this small timeline for this one reason. That refactor took a nosedive anyway, and that's because of the next reason:

1. The salvage+refactor approach produced huge unnavigable diffs that were impossible to audit and implanted very confusing history into git. We failed to push back on these earlier to allow the refactoring ahead of time. Why we defaulted to the no-fuss no-pushback approach that got us into that mess is the final point leading to the close:

2. **These are bug-fixes, and we want to ship them right away.** Like, tonight. It's not a feature that can sit in a PR while we meditate on what it needs or where it should go. We just need to fix the problem, and without causing another one while doing it. Ultimately we're responsible for what goes out each release _and_ we have to work with from everything that goes in from each commit.

So we are just going to fix the set of bugs as cleanly and comfortably as possible, and maybe if we're lucky, leave the immediate vicinity slightly nicer than when we found it. Please don't feel obligated to submit PR's along with issues thinking that it's going to lead to a faster fix, as if it relieves maintainers of another item to work through. The opposite might be true.

lhjt:
All good; I appreciate the time taken to look into these in addition to the explanation and response!

Next time I will await a response in the issues rather than immediately publishing a PR 👍🏻 (or is there something else you would rather me do?)

jevolk:
> All good; I appreciate the time taken to look into these in addition to the explanation and response!
> 
> Next time I will await a response in the issues rather than immediately publishing a PR 👍🏻 (or is there something else you would rather me do?)

I would say any PR following a bug issue should be as succinct and directed as possible, adding or modifying the absolute minimum to accomplish the objective of transitioning the state of `broken` -> `fixed` for that item.

PR #524, opened 2026-07-22
fix(api): clamp /messages pagination at the `to` position

basnijholt:
## Problem

`/messages` stopped pagination only when an event's count exactly equalled the `to` token (`Some(*count) != to`) and silently swallowed unparseable `to` values (`flat_ok()`). A sync `next_batch` token is a global stream position that almost never equals an event count *in the requested room*, so a `to` bound taken from `/sync` — which [the spec explicitly permits](https://spec.matrix.org/latest/client-server-api/#get_matrixclientv3roomsroomidmessages) — was ignored entirely and the walk ran unbounded.

Live probes against the lab server: `to=<sync token>` and even `to=garbage!!` both returned HTTP 200 with unclamped chunks.

Found by mindroom-nio's limited-timeline backfill (mindroom-ai/mindroom-nio#13), which pages `/messages` with sync tokens to recover events dropped by limited sync timelines. The nio client treats `to` as best-effort and does not depend on this fix — but with it, the server clamps the walk at the sync position, which also closes a federation topological-ordering edge case that client code cannot detect.

## Fix

- Stop the walk once it *passes* the `to` position in either direction (`count < to` forward, `count > to` backward), rather than only on exact count equality. Exact-equality events remain excluded, matching the previous behavior for the cases where it did work.
- Fail the request on an unparseable `to` (`transpose()?`), consistent with how an unparseable `from` fails, instead of silently ignoring it.

## Tests

New `src/mindroom-tests/tests/messages_to_clamp.rs` integration test drives the real router: it takes a sync `next_batch` while a second room advances the global stream (so the position falls strictly between the test room's events and equals none of their counts — the exact shape of the reported bug), then asserts the backward walk stops at the position, the forward walk starts at it (the two directions partition the room), and an unparseable `to` fails the request. The test fails against the previous behavior (backward walk runs past the position to room creation).

Full `mindroom_rebase_tests` suite passes; `cargo clippy -p tuwunel_api` is clean. `FORK_CHANGES.md` documents the change as fork change 6 — candidate for upstreaming.

### DIFF
--- a/.github/workflows/auto-mindroom-release.yml
+++ b/.github/workflows/auto-mindroom-release.yml
@@ -1,0 +1,92 @@
+name: Auto MindRoom Release
+
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  actions: write
+  contents: write
+
+concurrency:
+  group: auto-mindroom-release-main
+  cancel-in-progress: false
+
+jobs:
+  release:
+    name: Compute and create MindRoom release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4.2.0
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+
+      - name: Refresh remote tags
+        run: git fetch --force --tags origin
+
+      - name: Compute MindRoom release tag
+        id: version
+        env:
+          RELEASE_TAG_PREFIX: v
+          RELEASE_TAG_SUFFIX: mindroom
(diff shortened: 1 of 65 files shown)


basnijholt:
Opened against the wrong base (this targets the mindroom fork's tree, including fork-only files) — apologies for the noise. The underlying `/messages` `to`-clamping fix is upstream-relevant though; a clean upstream PR without the fork scaffolding can follow.

PR #541, opened 2026-08-07
core/config: Document the nine undocumented cache capacity fields.

byteflavour:
Nine cache capacity fields in `Config` (`pdu_cache_capacity`,
`auth_chain_cache_capacity`, `shorteventid_cache_capacity`,
`eventidshort_cache_capacity`, `eventid_pdu_cache_capacity`,
`shortstatekey_cache_capacity`, `statekeyshort_cache_capacity`,
`servernameevent_data_cache_capacity`, `stateinfo_cache_capacity`) only
carried `/// default: varies by system`, so `tuwunel-example.toml` renders
them as "This item is undocumented. Please contribute documentation for
it." Their neighbors `db_cache_capacity_mb` / `db_write_buffer_capacity_mb`
already explain what they're for; these nine didn't say anything.

I ran into this chasing an unrelated RAM report on my own homeserver and
ended up reading `cf_opts.rs` and `state_compressor/mod.rs` to find out
what each one actually sizes, since the fields alone don't say. This PR
writes that up: which RocksDB column family (or pair of column families)
each of the eight RocksDB-backed fields is bound to, what that column
family holds, and that the values are entry counts rather than byte
sizes -- the real cache size is entries times an internal per-column-family
key+value size hint, then `cache_capacity_modifier`, and neither of those
multipliers is something the operator sets directly. `stateinfo_cache_capacity`
turned out to be the odd one out: a plain in-process `LruCache`, no RocksDB
underneath it at all, sized straight in entries.

Docs only -- no `default: N` line changed, so `documented_defaults_match_the_code`
doesn't need updating.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,9 +19,9 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
+checksum = "c982642fa9e8606056828ee9a8505737230110bb1099153c79efe865c59d12ba"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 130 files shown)


jevolk:
Can you please set your branch to 92dc224117897804b854727aeed59262178540ae

byteflavour:
Done — branch is now at 92dc224. Thanks for regenerating the example config alongside it, that was missing from my original commit.

x86pup:
We'd rather not have the trailer, sorry.

byteflavour:
> We'd rather not have the trailer, sorry.

Okay i will respect this for future contributions to tuwunel. trailer is gone with 92dc224, all good.

jevolk:
Sorry about the state of this one; the mistake is entirely ours. Your change landed on main with your authorship intact as 64ff7b18, but a history rewrite on our side destroyed the merge commit that would have let GitHub record this pull request as merged, and the rewritten commit GitHub sees is no longer your branch head.

With "Allow edits by maintainers" unchecked we cannot repair the branch from our side, and the only other route would import stale duplicate history into main, so closing by hand is what remains. Ticking that box on future PRs gives us a recovery path when the fault is ours, as it is here. The work shipped exactly as you wrote it, and we appreciate the contribution.

PR #557, opened 2026-08-17
impl: recursive rocksdb multi get

gamesguru:
### What does this PR do?

Adds an optimized method which improves the speed of recursive queries (traversing deep forks can sometimes contribute to lag during peak federation hours). Puts an entire recursive query behind a single `tokio::spawnBlocking()` call and batches key lookups in RocksDB's monotonic order (for sequential reads and hot cache locality).

This PR does NOT hook the method up to anything (it's just there for future use). I don't have any explicit benchmark data yet, but it should be quite significant in many workloads.

### Checklist

      comment.
      Tuwunel's Code of Conduct.

---
## Summary by cubic
Adds `Map::recursive_multi_get` for breadth-first DAG traversal over `RocksDB`, returning parsed values, missing keys, and a truncation flag. Improves deep ancestor/fork queries via snapshot reads; not wired to production yet.

- Ordering and bounds: BFS with per-level key sort, root deduplication, and visited tracking; limits via `max_nodes`/`max_depth`. When `max_nodes` is hit mid-batch, the method now consumes the full multi-get batch to record missing keys and surface RocksDB errors without parsing additional values, and sets `truncated = true`.
- Consistency and performance: snapshot-based reads; batched multi-get with `SORTED=true`; pre-allocated child sink; per-level server shutdown checks; fail-fast error mapping via `util::map_err`.
- API and tests: exports `RecursiveGetOutput` from `map.rs` and `database/mod.rs`. Unit tests cover cycles, diamond convergence, missing keys, depth/node truncation, and mid-batch truncation behavior. Migration behavior unchanged (minor `Option::zip` refactor in `service/migrations/injectivity/repair.rs`).

<sup>Written for commit f249ba9009a4b7e1a678896d150cbceba660e996. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gamesguru/tuwunel/pull/8?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/database/map.rs
+++ b/src/database/map.rs
@@ -47,7 +47,10 @@ use self::watch::Watch;
 ///
 /// `Get` accepts raw keys, while `Qry` serializes structured keys before
 /// lookup. Both yield pinned value handles through an asynchronous stream.
-pub use self::{get_batch::Get, qry_batch::Qry};
+pub use self::{
+	get_batch::{Get, RecursiveGetOutput},
+	qry_batch::Qry,
+};
 use crate::{Engine, util::map_err};
 
 /// Provides typed and raw access to one RocksDB column family.
--- a/src/database/map/get_batch.rs
+++ b/src/database/map/get_batch.rs
@@ -189,6 +189,8 @@ where
 	let map = self.clone();
 
 	tokio::task::spawn_blocking(move || {
+		const SORTED: bool = true;
+
 		map.engine.ctx.server.check_running()?;
 
 		let snapshot = map.engine.db.snapshot();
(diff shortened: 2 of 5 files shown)


jevolk:
We already have a concurrent multi-threaded recursive traversal for e.g. deep auth-chain forks...

Why would we want the single-threaded sequential version with worst-case tail-latency?

gamesguru:
In the context of walking live DAGs, this method offers snapshot isolation and cache-friendly bulk reads.

It's possible we need to add intra-level parallelism to make this handle skinny/linear DAGs better.

Potential use cases include recursive relations and path finding (where `/backfill` fails or stalls indefinitely, e.g., M-HQ or Unrdct Lounge incidents).

jevolk:
> where /backfill fails or stalls indefinitely, e.g., M-HQ or Unrdct Lounge incidents

I have a strong suspicion you are conflating Tuwunel with other software. Can you file an issue which focuses on the problem and its reproduction rather than offer a questionable solution which we can't take as-is even if it were an effective solution.

We're also going to need exact details for the model, prompt, and base context you're using which came to this solution before we can meaningfully proceed on this.

PR #564, opened 2026-08-28
Send 0 notification counts unconditionally in sync v3

ruka-hamanasu:
### What does this PR do?

Previously, if a client updated its read receipt in the same sync round in which a non-notifying timeline event (reactions, edits) occurs, Tuwunel was not emitting a notification count of 0 (or any notification count). Because future syncs without further notifications would keep the notification count as 0, that means that any client relying only on seeing the 0 notification count (for example FluffyChat) would never reset its notification count and the room would remain marked as unread.

This PR includes both a shell script to reproduce the problem (see the instructions contained within) and the fix, with unit tests.

The code here is currently running on my homeserver (matrix.agiadn.org) along with #565, and along with some FluffyChat fixes (I'll link that PR when I put it up), FluffyChat is now marking rooms as read in all circumstances.

### Checklist

      comment.
- [n/a] Config option changes were made in `src/core/config/mod.rs` doc
- [n/a] User-facing changes are reflected in `docs/`.
      Tuwunel's Code of Conduct.


ruka-hamanasu:
CI is failing on changes I did not make, so I'm going to leave those failures for the maintainers to decide about.

jevolk (approved):

PR #576, opened 2026-09-06
tests: isolate default test databases

basnijholt:
### What does this PR do?

While checking upstream's test setup during the fork rebase, I noticed that `Args::default_test` selects `localhost` as the server name but leaves the database path to the normal configuration. Several smoke and listener tests use that helper without providing their own path. That matters because the `fresh` and `cleanup` hooks remove the selected database directory: those tests should not share a configured database or depend on each caller remembering an override.

This gives each call a randomly named database path under the system temporary directory. The path is stored in the arguments, so cloning or reapplying those arguments keeps the same database. An explicit `.with_option("database_path=...")` added afterward still wins, and ordinary server startup is unchanged. No new dependency or separate test crate is needed.

The regression tests check that a configured database path is overridden, concurrent callers get distinct paths, and explicit overrides still work. A database-lifecycle test also exercises the real `fresh` and `cleanup` hooks while a neighboring test directory holds a marker that must survive.

The fork already carries default test-database isolation; this extracts that behavior into upstream's existing argument builder and adds focused coverage there.

Verified in the pinned Nix dynamic shell (Rust 1.95.0 and nightly rustfmt):

- `cargo fmt --all -- --check`
- `cargo clippy --offline --locked --workspace --all-targets --all-features -- -D warnings`
- `cargo test --offline --locked --workspace --all-targets --all-features` — 1,082 passed, 0 failed, 4 ignored, matching upstream's ignored count.

I first ran the regression against unchanged upstream and confirmed the missing/inherited-path failures. I also ran the final integration-test binary in three simultaneous processes sharing a temporary directory with spaces and quotes: all 12 checks passed, and no test databases remained. Test runs used isolated networking and storage, with debug symbols disabled.

### Checklist

- [x] Nightly formatting and pinned stable Clippy/rustc checks pass; intentional lint expectations explain why quoting and exclusive directory creation are needed.
- [x] Complement was not run locally; no compliance-result changes are claimed.
- [x] No public configuration options changed, so no example-config regeneration is needed.
- [x] The test-helper behavior is documented in source; no operator-facing feature was added.
- [x] I agree to the Apache-2.0 licensing terms and the project's Code of Conduct.

### DIFF
--- a/src/main/args.rs
+++ b/src/main/args.rs
@@ -1,13 +1,17 @@
 //! Integration with `clap`
 
-use std::{env::var, path::PathBuf, process::id as process_id};
+use std::{
+	env::{temp_dir, var},
+	path::PathBuf,
+	process::id as process_id,
+};
 
 use clap::{ArgAction, Parser, builder::RangedU64ValueParser};
 use tuwunel_core::{
 	Err, Result,
 	config::{Figment, FigmentValue},
 	err, implement, is_true, toml,
-	utils::available_parallelism,
+	utils::{available_parallelism, random_string},
 };
 
 /// Only its own argument may set this, since restoring is destructive and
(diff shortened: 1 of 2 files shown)


basnijholt:
I checked the [Format failure](https://github.com/matrix-construct/tuwunel/actions/runs/34047277850/job/101524500226). It reports comment reflow in 130 files already present on `main`. The new `src/main/tests/default_test_database.rs` is not flagged; the two hunks in `src/main/args.rs` are unchanged comments in `update`, outside this patch.

The pinned nightly formatter passes locally. I'm keeping that repository-wide reformat out of this PR.

basnijholt:
I reproduced the current nightly formatting and Clippy failures on unmodified `main`. The shared fix is in #578, so I've kept that cleanup out of this diff. The flagged service files are unchanged here. Once #578 lands, this PR can be rebased onto it and CI rerun.

basnijholt:
@jevolk I fixed the merge conflicts here and rebased.

PR #577, opened 2026-09-06
api/client: preserve device keys on read errors

basnijholt:
### What does this PR do?

While reviewing the remaining device-key changes in the fork, I noticed that the exact-copy check in `/keys/upload` also hides errors reading the existing keys. The `is_ok_and` chain turns a database or decoding error into `false`, then the handler writes the uploaded keys over the stored row. A failed read is not evidence that no keys exist or that they differ.

This distinguishes a missing row from a failed read. A genuinely missing row still permits the first upload; other read or decoding errors return a server error and leave the stored device-key bytes untouched. Malformed keys in the request are still rejected as a client error before this check.

The successful cases stay the same: uploading different valid device keys still replaces the existing keys, and an exact-copy retry is still ignored so it cannot strip the existing cross-signing signatures. The fork has a stricter identity-replacement policy, but that policy is not part of this PR.

The regression uses the existing server-booting HTTP harness with a real database. It checks first upload, valid replacement, and signature-preserving retries, then inserts malformed stored rows: missing required fields, invalid JSON syntax, truncated JSON, and invalid UTF-8. Each corrupt-row upload must return HTTP 500 and preserve the original bytes. These fixtures exercise decoding failures; they do not simulate physical storage I/O failures.

Verified in the pinned Nix dynamic shell (Rust 1.95.0 and nightly rustfmt):

- `cargo fmt --all -- --check`
- `cargo clippy --offline --locked --workspace --all-targets --all-features -- -D warnings`
- `cargo test --offline --locked -p tuwunel --all-features --test device_key_read_errors` — passed.
- `cargo test --offline --locked --workspace --all-targets --all-features` — 1,079 passed, 0 failed, 4 ignored, matching upstream's ignored count.

I also ran the regression against unchanged upstream: all four corrupt stored rows returned HTTP 200 and were overwritten. Test runs used isolated networking and storage, with debug symbols disabled.

### Checklist

- [x] Nightly formatting and pinned stable Clippy/rustc checks pass; the unused shared-harness helpers have a scoped, reasoned lint expectation.
- [x] Complement was not run locally; no compliance-result changes are claimed.
- [x] No public configuration options changed, so no example-config regeneration is needed.
- [x] This fixes existing endpoint error handling without adding an operator-facing feature or setup requirement.
- [x] I agree to the Apache-2.0 licensing terms and the project's Code of Conduct.

### DIFF
--- a/src/api/client/keys/upload_keys.rs
+++ b/src/api/client/keys/upload_keys.rs
@@ -2,7 +2,7 @@ use axum::extract::State;
 use ruma::{
 	DeviceId, UserId, api::client::keys::upload_keys, encryption::DeviceKeys, serde::Raw,
 };
-use tuwunel_core::{Err, Result, debug, err};
+use tuwunel_core::{Err, Result, debug, err, utils::result::NotFound};
 use tuwunel_service::Services;
 
 use crate::Ruma;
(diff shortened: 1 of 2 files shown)


basnijholt:
I reproduced the current nightly formatting and Clippy failures on unmodified `main`. The shared fix is in #578, so I've kept that cleanup out of this diff. The flagged service files are unchanged here. Once #578 lands, this PR can be rebased onto it and CI rerun.

PR #578, opened 2026-09-06
chore: fix current nightly lints

basnijholt:
While checking CI on #576 and #577, I reproduced the nightly formatting and Clippy failures on unmodified current `main`. They are shared baseline failures, so I'm keeping the cleanup in its own PR instead of adding it to either feature patch.

This sets `wrap_comments = false` in `rustfmt.toml`, so the new nightly formatter leaves the existing comment layout alone. The original comment reflow has been removed from the diff; there are now nine changed files.

The remaining source changes simplify five imports ending in `::{self}` and remove six clones flagged by nightly Clippy. Each removed clone was at the value's final use. The clones still needed across state-resolution iterations and for the sending transaction status map remain in place. No toolchain pin or new lint allowances are added.

Verified locally on aarch64 Linux after this revision:

- `cargo fmt --all -- --check` with the formatter from the pinned development shell, nightly-2026-09-05, and nightly-2026-09-06.
- Clippy with `-D warnings` across all workspace targets and all features: Rust 1.95.0, nightly-2026-09-05, and nightly-2026-09-06. Both nightly runs use CI's test profile and existing solver/unstable-feature flags.
- Rust 1.95.0 `cargo test --offline --locked --workspace --all-targets --all-features`: 1,078 passed, zero failed, four existing ignored tests; benchmark smoke checks passed too. Server tests used isolated local databases and loopback networking.
- All original comment lines and the generated `tuwunel-example.toml` are unchanged from `main`.

I haven't run Complement locally; no protocol or compliance change is intended.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -33,4 +33,4 @@ unstable_features = true
 use_field_init_shorthand = true
 use_small_heuristics = "Off"
 use_try_shorthand = true
-wrap_comments = true
+wrap_comments = false
--- a/src/api/client/membership/members.rs
+++ b/src/api/client/membership/members.rs
@@ -2,7 +2,7 @@ use axum::extract::State;
 use futures::{FutureExt, StreamExt, pin_mut};
 use ruma::{
 	api::client::membership::{
-		get_member_events::{self},
+		get_member_events,
 		joined_members::{self, v3::RoomMember},
 	},
 	events::{
--- a/src/api/client/send.rs
+++ b/src/api/client/send.rs
@@ -17,8 +17,7 @@ use serde_json::from_str;
 use tuwunel_core::{
 	Err, Result, debug_warn, err,
 	matrix::{Event, pdu::PduBuilder},
-	utils::{self},
-	warn,
+	utils, warn,
 };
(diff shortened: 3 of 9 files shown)


jevolk:
Hi @basnijholt, it looks like `rustfmt` released some breaking changes a few hours ago. Since this is not a regression we're not going to pin an older nightly.

Instead of reformatting 137 files, try setting `wrap_comments = false` in the rustfmt.toml. This PR should then be reduced to your `use` statement changes with no other noise.

Sorry about our bleeding-edge lifestyle here, but it does have its merits.

basnijholt:
Set `wrap_comments = false` and restored the original comment layout. The diff is now nine files, with no comment reflow.

I kept the six redundant-clone fixes alongside the five import cleanups because those clones also cause fatal errors in the nightly Clippy jobs. The required clones are unchanged.

Rebuilt this as one commit and force-pushed. Fresh formatting checks pass with the pinned development formatter and both September 5 and September 6 nightlies; stable and nightly Clippy pass too. The isolated test run passed 1,078 tests, with four existing ignored tests.

jevolk (approved):

PR #580, opened 2026-09-08
Add configurable language for URL previews

### LINKED ISSUE #575: Add Accept-Language header config
### What problem would this solve?

I host my server with Hetzner in Germany, and as a result URL previews often return their results in German.

### What would you like to happen?

Add a configuration alongside the user-agent configuration where you can set a preferred Accept-Language header so you can specify English as a result. While this still won't stop sites that only use geolocation via IP address it would help with sites that honor Accept-Language.

### Alternatives or workarounds

_No response_

### Related MSC or spec section

_No response_

tototomate123:
### What does this PR do? 
URL previews can appear in the language associated with the homeserver's IP location. This adds an optional `url_preview_accept_language` configuration option so administrators can request their preferred language from websites that honor HTTP language negotiation.

For example:
```toml
url_preview_accept_language = "en-US,en;q=0.9"
```
The setting applies to URL-preview pages, media, and oEmbed requests, and takes effect for new requests after the configuration is reloaded. When unset, no header is sent, while invalid HTTP header values are rejected during startup or reload validation; the setting is also documented in the configuration comments and generated example file.

Existing cached previews retain their current language until they expire.
Websites that select language only by IP address may ignore this header.

Fixes #575 

### Testing
- `cargo +1.95.0 test -p tuwunel_core url_preview_accept_language --locked`
=> 3 tests passed
- `cargo +1.95.0 test -p tuwunel_core config::regenerate::tests::pristine_output_matches_the_checked_in_example --locked`
=> passed
- `cargo +1.95.0 check -p tuwunel_service --features url_preview --locked`
=> passed

### Checklist

- [x] Code is formatted with nightly cargo fmt and satisfies clippy and rustc lints; any allowed lint is justified by an obvious reason or a comment
- [x] No Complement compliance changes were made; this only affects URL preview requerst configuration
- [x] The url_preview_accept_language option is documented in src/core/config/mod.rs, and the regenerated tuwunel-example.toml is commited
- [x] No seperated docs/ page was needed
- [x] I agree that my changes may be licensed under the Apache-2.0 licence and my conduct is in line with the Contributor's Covenant and Tuwunel's Code of Conduct.


tototomate123:
Force pushed a CI fix that explicitly imports `config::check::reload` and removes redundant qualifications throughout the config tests - addresses the unused qualifications failure under warnings as errors

jevolk (approved):

PR #581, opened 2026-09-08
feat(policy): mandatory policy server + fail-closed mode

andrimuhyidin:
Adds four opt-in `[global]` knobs. With them unset, behaviour is byte-for-byte upstream.

| knob | meaning |
|---|---|
| `policy_server_mandatory_url` | `/sign` target for **every** local event, bypassing the federation resolver, `allow_federation`, the denylist, and the "policy server must have a user in the room" requirement |
| `policy_server_mandatory_name` | name used in the `signatures[...]` slot and the X-Matrix `destination` |
| `policy_server_mandatory_public_key` | ed25519 key used to verify the signature that comes back |
| `policy_server_fail_closed` | policy server unreachable / timing out / in backoff → **reject** the send (default `false`, i.e. today's behaviour) |

The three `mandatory_*` knobs are all-or-nothing and require `enable_policy_servers`; `config/check.rs` enforces that.

## Why

Deployments that must guarantee a policy decision on every event cannot rely on `m.room.policy`: it is room state, so a room creator can remove it. In v12 rooms that removal is itself not policed, so the guard disappears **silently**.

Fail-open is also the wrong default for those deployments — an unreachable policy server currently means events flow unchecked, which is indistinguishable from a policy that allows everything.

## Implementation

- `PolicyTarget` + `lookup_policy_target` in `policy_server.rs` — the mandatory target short-circuits the resolver.
- `federation.execute_at` in `federation/execute.rs` — sends to an explicit URL/name pair instead of resolving a server name.
- `react_to_fetch` is now pure, with table tests.
- `m.room.create` is evaluated too. The room version is read from the create event's own content; without that the existing guard skips it silently — exactly the hole this mode exists to close.

`tuwunel-example.toml` is regenerated by the proc-macro at compile time, not hand-edited.

## Testing

`cargo +nightly fmt --check` and `cargo clippy --workspace --all-targets` are clean.

Behaviour is proven end-to-end against a real tuwunel binary and a stub policy server — not mocked:

- every local event calls `/sign`, including `m.room.create`
- rejections come back as 403 `M_FORBIDDEN`
- `/sign` carries X-Matrix with `destination` set to the configured name
- a v12 room whose creator **removes** `m.room.policy` is still policed
- with `fail_closed = true`, a policy server returning 500 — or entirely down — rejects the send, and the warning names `fail_closed`
- **control:** with `fail_closed = false` the same dead server lets the event through, matching current behaviour

Happy to adjust naming, split the commit, or move the fail-closed knob under a different section if you prefer.

### DIFF
--- a/docs/development/compliance/msc.md
+++ b/docs/development/compliance/msc.md
@@ -63,7 +63,7 @@ in the [Out of scope](#out-of-scope) section.
 | MSC4297 | ✅ ● | 100/100 | State Resolution v2.1 | src/service/rooms/state_res/resolve.rs:257 conflicted state subgraph; tests pass |
 | MSC4291 | ✅ ● | 100/100 | Room IDs as hashes of the create event | v12 upgrade create event omits deprecated predecessor.event_id |
 | MSC4289 | ✅ ● | 100/100 | Explicitly privilege room creators | Room v12 creator privilege; complement 22 pass, 0 fail across six named tests |
-| MSC4284 | 🟨 ● | 70/90 | Policy Servers | outbound /sign, inbound verify, fetch-on-missing, reversible soft-fail |
+| MSC4284 | 🟨 ● | 70/90 | Policy Servers | outbound /sign, inbound verify, fetch-on-missing, reversible soft-fail; operator-mandatory server at a direct URL and optional fail-closed (beyond the MSC) |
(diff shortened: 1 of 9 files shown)


jevolk:
This has a few issues that will prevent intuitive operation, but it can be refactored a bit to add some value to the existing policy server spec. Please don't close this since it looks like a good starting point.

PR #588, opened 2026-09-12
federation: allow omitted x-matrix destination parameter for compatibility

### LINKED ISSUE #123: High CPU and Read-IO on main branch
Observations from real-world usage of the project hydra updates deployed to the main branch show abnormally sustained CPU and disk-read of an unknown origin.

kybe236:
See https://spec.matrix.org/v1.19/server-server-api/#request-authentication

### What does this PR do?

Allows federation from older servers which do not send the destination field in the X-matrix header like the spec suggests.
image

Also changes the error code to the one recommended in the spec (403 -> 401).

Changes where made due to me noticing blocked federation resulting in constant 403 errors in my metrics which led to me looking into the requests and responses.

image

image

### Checklist

- [X] Code is formatted with nightly `cargo fmt` and satisfies clippy and
      comment.
- [X] Complement compliance changes (new passes or new failures), if any,
- [X] I agree that my changes may be licensed under the Apache-2.0 licence
      Tuwunel's Code of Conduct.

### DIFF
--- a/src/api/router/auth/server.rs
+++ b/src/api/router/auth/server.rs
@@ -96,8 +96,9 @@ fn auth_server_checks(services: &Services, x_matrix: &XMatrix) -> Result {
 	}
 
 	let destination = services.globals.server_name();
-	if x_matrix.destination.as_deref() != Some(destination) {
-		return Err!(Request(Forbidden("Invalid destination.")));
+	let x_matrix_destination = x_matrix.destination.as_deref();
+	if x_matrix_destination.is_some() && x_matrix_destination != Some(destination) {
+		return Err!(Request(Unauthorized("Invalid destination.")));
 	}
 
 	let origin = &x_matrix.origin;


x86pup:
I don't think this should be forced. We aren't the first server to enforce it, it goes all the way back to Conduit and there has never been a problem. Every single server is compatible with this change and there has not been any issue.

What is the actual usecase here? You say you have a server that doesn't handle this correctly. Why?

Also, touching the error code might break other servers.

I think this PR needs a ton of research and validation before touching this sensitive area here.

kybe236:
The practical impact is that servers running older Synapse versions (like 1.56.0) get blocked. As the spec notes, recipients should accept requests without this parameter for compatibility, so this PR brings Tuwunel closer to that recommended behavior and aligns with [Synapse's own implementation](https://github.com/element-hq/synapse/blob/3f17bd706ae9043e31db3e9d24a259ca42c9a312/synapse/federation/transport/server/_base.py#L110-L118). If you prefer to keep the legacy Conduit strictness despite that, feel free to close the PR.

jevolk:
I looked into this and there is no reason to be strict. I also despise `destination` in general because it inhibits zero-copy broadcasts to multiple servers. I also looked into the 401 vs. 403 and Synapse handles the 401 better for this response.

Issue #282, opened 2026-02-01
Implement rate limits mandated by the spec


grinapo:
(Opening issue by proxy for Emma)

There are various points where the spec requests an API to be rate limited.

If I see correctly S2S only requests `GET /_matrix/federation/v1/media/*` to be rate limited.

However plenty of C2S API calls should be rate limited, which is important for various security and abuse management purposes. These shall be implemented. 
Probably some earlier than the others…

PR #597, opened 2026-09-20
login_ratelimit: Limit password login attempts per account.

obodnikov:
### What does this PR do?

Bounds password guessing, one of the rate limits the spec requires and #282 collects.

`/_matrix/client/v3/login` with `m.login.password` counts nothing today: the password path looks up the hash and verifies it, and nothing delays or refuses after any number of wrong answers. The only always-on guess limit in the tree is `check_device_rate_limit` on the OIDC device-code endpoints, and the optional `oidc_rc_*` throttle covers the OIDC endpoints only. The spec marks `/login` "Rate-limited: Yes".

**Two token buckets per account**, in a service of their own, taking Synapse's `rc_login.account` and `rc_login.failed_attempts` defaults unchanged:

| Option | Default | Synapse |
|---|---|---|
| `login_rc_account_per_second` | 0.003 | `rc_login.account.per_second` |
| `login_rc_account_burst_count` | 5 | `rc_login.account.burst_count` |
| `login_rc_failed_per_second` | 0.17 | `rc_login.failed_attempts.per_second` |
| `login_rc_failed_burst_count` | 3 | `rc_login.failed_attempts.burst_count` |

Every password attempt debits the account bucket; only a wrong password debits the failed bucket, which is consulted before the check, so a drained one refuses even a correct password. `/login` and `/_tuwunel/oidc/native` check both; UIAA password re-entry checks the failed axis only, since its caller already holds an access token but a stolen token must not become an unlimited password oracle. A burst of `0`, or a rate of `0`, disables a bucket.

**The account axis is stricter than Synapse's, on purpose.** Synapse checks `rc_login.account` in `_complete_login`, after the password has been verified, so there it counts completed logins, and guessing is left to `rc_login.failed_attempts` and `rc_login.address`. Here there is no address axis to lean on (next paragraph), and the failed bucket alone would still let one account be guessed at about ten attempts a minute from any number of addresses. So every password attempt debits the account bucket, which bounds guessing to its refill rate: after the burst, one attempt about every five and a half minutes.

**Keyed on the account, not on the client address.** A homeserver behind a reverse proxy or tunnel that does not forward the caller's address sees every request arrive from one IP; a per-address limit there either never trips or throttles every user at once. A per-account limit bounds guessing against one account whatever address the guesses come from. Unknown accounts are throttled exactly like real ones, so the `429` is not an oracle for which accounts exist.

**Why `service/login_ratelimit/` and not a couple of fields on `users::Service`.** That was the first shape, and it made clippy report an unfulfilled `#[expect(clippy::must_use_candidate)]` in `service/users/device.rs`, a file this PR does not touch: a `Mutex` field gives the type interior mutability, so clippy stops treating its `&self` methods as taking nothing mutable and the expectation no longer holds. A service of its own keeps that lint where it was, and matches how `threepid` already holds its own buckets.

**What this costs, stated plainly.** While somebody keeps guessing an account's password, its owner cannot sign in *with that password* either — the refusal is what makes the limit a limit. Nothing is signed out, and no path that is not a password is counted: existing sessions, `m.login.token`, SSO and JWT are untouched. The lock lasts exactly as long as the guessing. Synapse makes the same trade on its failed-attempts axis with the same defaults; here the account axis makes it too, which is the price of having no address axis.

**The table is bounded, and a spray cannot reset a victim.** Each bucket table holds at most 65,536 accounts. Past that, a new account is admitted only by pruning a bucket that has refilled completely; **a bucket that is still limiting is never evicted**, because evicting it is precisely what a spray of distinct user names would want. When nothing can be pruned the table fails closed for accounts it does not yet hold, and logs a warning at most once a minute. Pruning inspects a bounded sample rather than the whole table, so a full table costs a login constant work under the lock instead of an O(n) scan. `retry_after` is capped at a day so no configured rate can overflow the conversion.

**Complement:** the suite signs one account in many times over and tests wrong passwords on purpose, so `nix/pkgs/complement/config.toml` disables both limits, as other homeservers' test images do. No pass or failure should move. I have not run the suite myself; that line is from reading its configuration.

**Integration tests:** `oidc_completion` signs one account in sixteen times through the native login page, so it now switches the account axis off with `login_rc_account_per_second=0`, beside its existing `oidc_rc_per_second=0`. That test was the red job on this PR's first CI run.

**Testing.** 11 unit tests cover the burst, the refill, the peek-then-debit gate, case folding of the account key, the disabled settings, the capped `retry_after`, and the full-table behaviour including the spray that must not reset a victim's bucket. At this commit, rebased onto `main` at v1.9.2, on macOS (aarch64): nightly `cargo fmt --check` is clean; `cargo clippy --workspace --all-targets --no-deps -- -D warnings -A unstable-features` is clean; `cargo test --workspace --lib` passes, 574 of it in `tuwunel_service`; and `cargo test --test '*'` passes on every target except `state_local_build`, which fails on macOS identically on `main` at v1.9.2 without this change (`EAGAIN` in its `direct-memo-failure` case) and passes in your Linux CI. The features were the subset this machine builds — `brotli_compression,element_hacks,gzip_compression,media_thumbnail,url_preview,zstd_compression,ldap,direct_tls`; io_uring, systemd and jemalloc were not built here. Measured against a live server reached through a tunnel, so the server saw one address for every request: three wrong passwords answered `403`; the fourth answered `429 M_LIMIT_EXCEEDED` with `retry_after_ms: 5000`, and so did a correct password until the bucket refilled seven seconds later; past the burst of five the account axis took over with about five minutes; throughout, an access token issued earlier kept answering `whoami` and `/sync`.

Part of #282.

### Checklist

      comment.
- [x] User-facing changes are reflected in `docs/`.
      Tuwunel's Code of Conduct.


obodnikov:
Rebased onto `main` at v1.9.2 and fixed the red integration jobs. `oidc_completion` signs one account in sixteen times through the native login page, so the new account limit refused the sixth login with a `429`; the test now switches that axis off with `login_rc_account_per_second=0`, beside its existing `oidc_rc_per_second=0`.

I've also corrected the description where it said the account bucket mirrors Synapse's `rc_login.account`: Synapse counts only completed logins against it, and here every attempt counts, because there is no address axis to bound guessing. The reasoning is in the new paragraph after the table.

On macOS the whole integration suite passes except `state_local_build`, which fails there on `main` too.

jevolk:
This is not being ignored. I am currently away and June is working on hotfixes for 1.9.2. We may not get to this until next week or so.

